PR #10, opened 2022-11-05
Switch to user-runtime config approach

### LINKED ISSUE #9: Not working with Credentials provider
### Environment

Nuxi 3.0.0-rc.11
Nuxt 3.0.0-rc.11 with Nitro 0.5.4
@sidebase/nuxt-auth 0.0.1-beta.3

### Reproduction

_No response_

### Describe the bug

Thank you for this awesome module, it is the missing piece for the Nuxt universe.

I have try to setup nuxt-auth with the credentials Provider which is the most common type when using a decoupled backend.

this is my auth config in nuxt.config.ts copied from official docs in NextAuth

NUXT_URL =  http://localhost/api/auth/

```
  auth: {
    nextAuth: {
      url: process.env.NUXT_URL,
      options: {
        secret: process.env.NUXT_SECRET,
        providers: [
          CredentialsProvider({
            // The name to display on the sign in form (e.g. 'Sign in with...')
            name: 'Credentials',
            // The credentials is used to generate a suitable form on the sign in page.
            // You can specify whatever fields you are expecting to be submitted.
            // e.g. domain, username, password, 2FA token, etc.
            // You can pass any HTML attribute to the <input> tag through the object.
            credentials: {
              username: { label: 'Username', type: 'text', placeholder: 'jsmith' },
              password: { label: 'Password', type: 'password' },
            },
            async authorize(credentials, req) {
              // You need to provide your own logic here that takes the credentials
              // submitted and returns either a object representing a user or value
              // that is false/null if the credentials are invalid.
              // e.g. return { id: 1, name: 'J Smith', email: 'jsmith@example.com' }
              // You can also use the `req` object to obtain additional parameters
              // (i.e., the request IP address)
              // eslint-disable-next-line no-console
              console.log(credentials)
              const user = { id: '1', name: 'J Smith', email: 'jsmith@example.com' }

              if (user) {
                // Any object returned will be saved in `user` property of the JWT
                return user
              }
              else {
                // If you return null then an error will be displayed advising the user to check their details.
                return null

                // You can also Reject this callback with an Error thus the user will be sent to the error page with the error message as a query parameter
              }
            },
          }),
        ],
      },
    },
```

Login form its showing but not matter what is the input for username and password it always show the error: "Sign in failed. Check the details you provided are correct."

the authorize function its never reach when submitting the form.

Is there anything else I am missing in order to make it work

### Additional context

_No response_

### Logs

_No response_

BracketJohn:
Contributes to #9

This PR switches our auth-endpoint setup approach to one where the user does it at runtime. We then provide the helpers for it, but the user creates the module with config.

This has some advantages:
- follows nextauth convention,
- all imports are made by the user at runtime, so no more de-/serialization of options required

It has the disadvantage of forcing the user to do more work.

_This PR is WIP and untested, but the implementation is very easy based on our existing work, so it should hopefully not be much more to make this work._ Once it's done I would rip out all of the code that is only there to support the "old" approach, that should also simplify the module quite a bit.

Checklist:


BracketJohn:
Status:
- impl done
- open:
  - [ ] update docs
  - [ ] check out `useSession.ts` credential related todos -> they may apply here now
  - [ ] fix tests

PR #15, opened 2022-11-09
docs: improve readme

Atinux:
- `nuxt` -> `Nuxt`
- `middlewares` -> `middleware` (https://www.wordhippo.com/what-is/the-plural-of/middleware.html#:~:text=The%20noun%20middleware%20can%20be,or%20a%20collection%20of%20middlewares.)
- client-side middleware -> application-side middleware (this they are also called on server-side during SSR)
- Removed imports that are auto-imported

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@
         await signOut() // Sign out the user
         ```
 
-There's more supported methods in the `useSession` composable, you can create client- and server-side authentication middlewares for your app and more - read the documentation below.
+There's more supported methods in the `useSession` composable, you can create APP and API authentication middleware - read the documentation below.
 
 ## Features
 
@@ -64,7 +64,7 @@ There's more supported methods in the `useSession` composable, you can create cl
     - `useSession` composable to: `signIn`, `signOut`, `getCsrfToken`, `getProviders`, `getSession`
     - full typescript support for all methods and property
 - ✔️ Persistent sessions across requests
-- ✔️ Client-side middleware protection
+- ✔️ Application-side middleware protection
 - ✔️ Server-side middleware and endpoint protection
 - ✔️ REST API:
     - `GET /signin`,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,7 +51,7 @@
         await signOut() // Sign out the user
         ```
 
-There's more supported methods in the `useSession` composable, you can create client- and server-side authentication middlewares for your app and more - read the documentation below.
+There's more supported methods in the `useSession` composable, you can create APP and API authentication middleware - read the documentation below.
```
BracketJohn, line 54: Two things:
- I'd either use `App` or `Application` here -> you do the same below
- Where is it established that "App" is the right name for what we called client-side before, I'm not aware of App meaning: Client-/Browser-side behavior

README.md
```diff
@@ -102,16 +102,16 @@ Below we describe:
         - [basePath](#basepath)
     - [NuxtAuthHandler](#nuxtauthhandler)
         - [Example with two providers](#example-with-two-providers)
-2. [Client-side usage](#client-side-usage)
+2. [Applicaiton-side usage](#application-side-usage)
```
BracketJohn, line 105:
```suggestion
2. [Application-side usage](#application-side-usage)
```

BracketJohn (changes requested):

BracketJohn:
Thanks for your contribution @Atinux - it's largely good, and we should be able to merge it soon ❤️

README.md
Atinux: - Then use Application
- You need to understand that the middleware/ directory is also called when server-rendering the page, this is why using client-side is wrong IMO

BracketJohn: Alright, then let's use Application.

I get it, another idea that may be clearer is to call them "universal- and server-middlewares"? This is just a nit though, so you decide which way to go here. 

If you're fine with any I may merge the PR later and then do the xhange myself to unblock you here / require less of your attention.

BracketJohn (approved):

PR #40, opened 2022-11-17
feat!: switch session interface to have `.data` dependent on `.status`

BracketJohn:
This PR improves the typing of the `useSession` composable, either:
- `session.value.status === 'authenticated'` and `session.value.data` will be available
- `session.value.status !== 'authenticated'` and `session.value.data` will not be available, resulting in a TS error

This will force users of `nuxt-auth` to always check if the use is `authenticated` before relying on `.data`. Non TS users will not be impacted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,13 +39,13 @@
 4. Done! You can now use all user-related functionality, for example:
     - application-side (e.g., from `.vue` files):
         ```ts
-        const { status, data, signIn, signOut } = await useSession({
+        const { session, signIn, signOut } = await useSession({
           // Whether a session is required. If it is, a redirect to the signin page will happen if no active session exists
           required: true
         })
 
-        status.value // Session status: `unauthenticated`, `loading`, `authenticated`
-        data.value // Session data, e.g., expiration, user.email, ...
+        session.value.status // Session status: `unauthenticated`, `loading`, `authenticated`
+        session.value.data // Session data, e.g., expiration, user.email, ...
 
         await signIn() // Sign in the user
         await signOut() // Sign out the user
@@ -118,6 +118,7 @@ Below we describe:
     - [Session access and manipulation](#session-access-and-manipulation)
         - [Redirects](#redirects)
         - [Custom sign in page](#custom-sign-in-page)
+        - [The session interface](#the-session-interface)
     - [Middleware](#middleware)
         - [Global middleware](#global-middleware)
         - [Named middleware](#named-middleware)
@@ -274,8 +275,7 @@ The `useSession` composable is your main gateway to accessing and manipulating s
 
 ```ts
 const {
-  status,
-  data,
+  session,
   getCsrfToken,
   getProviders,
   getSession,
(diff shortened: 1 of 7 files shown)


BracketJohn:
@valiafetisov my main interest would be: How do you like this design choice? I feel that it would be a great addition for people using TS and will not impact people using JS + it's a change we can still make in this stage of module-popularity. A possible downside is, that this deviates from `NextAuth.js` data-models, [see here](https://next-auth.js.org/getting-started/client#usesession)

valiafetisov:
I like the addition, but why type enforcing is not possible with the previous structure? I can imagine that `{ status, data } = await useSession()` can be related in the same way as `{ status, data } = session` types.

> A possible downside is, that this deviates from NextAuth.js data-models, [see here](https://next-auth.js.org/getting-started/client#usesession)

Thanks for the context I was missing. Maybe then it's not such a good idea after all.

BracketJohn:
Well, I couldn't come up with a way to make the two related as independent variables. And it turns out that TS in general cannot keep track of refinements (eg, what happens when you check for one variant of a discriminate union): https://kyleshevlin.com/discriminated-unions-and-destructuring-in-typescript

---

Wheb you're saying maybe, do you have a preference going any one way? (:

noook:
I don't think it's possible to do it this way. I think the issue is if you destructure into multiple variables, Typescript is not able to link them anymore.

It should look like that if you're trying to achieve a discriminated union:

```ts
const session = await useSession();
if (session.status.value === 'authenticated') {
  session.data.value;
  // ^? -> Session
}
session.data.value;
// ^? -> Session | null
```

BracketJohn:
This, or actually a stricter version of this, is what's implemented in this PR at the moment. Atm. this code produces the following:
```ts
const { session } = await useSession({ required: false })
if (session.value.status === 'authenticated') {
  session.value.data;
  // ^? -> Session
} else {
  session.value.data;
  // ^? -> TS Error: `data` does not exist on session.value
}
```

> I think the issue is if you destructure into multiple variables, Typescript is not able to link them anymore.

I thought the same, but as per https://stackoverflow.com/a/69033664 there is a way to make it work now (I've not tried it yet)

valiafetisov:
> And it turns out that TS in general cannot keep track of refinements

...after destructuring. Cause in general it can

> This, or actually a stricter version of this, is what's implemented in this PR at the moment

I thought that the [comment above](https://github.com/sidebase/nuxt-auth/pull/40#issuecomment-1321188203) suggests that in order to enforce types you not only need to set correct union types, but also use variables without destructuring. I.e. update our readme examples:

```diff
- const { status, data, signIn, signOut } = await useSession()
+ const session = await useSession()
```

Or wait till typescript will support it.

> Wheb you're saying maybe, do you have a preference going any one way? (:

I'm not sure it's a good idea to introduce breaking changes just to overcome some ts limitation (that might be fixed in the future and can already be overcome by using different syntax). Otherwise, someone might still write `const { status, data } = session.value` and loose type enforcing again.

BracketJohn:
closing this for now as outdated, we can revisit it later if we want to (:

PR #51, opened 2022-11-24
Automatically infer port in dev-mode

### LINKED ISSUE #50: Use nuxt nitro config to set host and port in local development
### Describe the feature

Currently in dev-mode we set the origin for the user unless they set their own `origin` using the module config:
https://github.com/sidebase/nuxt-auth/blob/44a159c5d936869c3c9d5e41d337a1578174bec1/src/module.ts#L61-L62

This leads to problems when users have their app running during development on a non-standard port (or host) during development. It does not lead to a production problem as we enforce a manually set `origin` during production for security reasons.

On discord:
- a diuscord user  @ThatTonybo actually faces this problem [1]
- @danielroe provided a solution that (at the moment) works to determine host and port during development: https://discord.com/channels/473401852243869706/473406398005248011/1044975921033130088

Let's apply the solution as discussed on discord. We should ensure a fallback to `3000` in case nothing is set:
```
nuxt.options.server.port || process.env.PORT || process.env.NITRO_PORT || 3000
```

Let's: 
- [ ] make port-detection dynamic
- [ ] make host-detection dynamic

[1]
image

### Additional information

_No response_

IronicUsername:
Closes #50

### Description
- make it possible to set a different port using either
  - `nuxt` options
  - `PORT` env var or
  - `NITRO_PORT` env var
  - fallback is `3000`
- adjust `README` accordingly 
- fix 1 smol typo (`npm run type` -> `npm run types`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,6 @@
 [![Follow us on Twitter](https://badgen.net/badge/icon/twitter?icon=twitter&label)](https://twitter.com/sidebase_io)
 [![Join our Discord](https://badgen.net/badge/icon/discord?icon=discord&label)](https://discord.gg/9MUHR8WT9B)
 
-
 > Nuxt user authentication and sessions via [NextAuth.js](https://github.com/nextauthjs/next-auth). `nuxt-auth` wraps [NextAuth.js](https://github.com/nextauthjs/next-auth) to offer the reliability & convenience of a 12k star library to the nuxt 3 ecosystem with a native developer experience (DX).
 
 ## Quick Start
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -19,74 +18,88 @@
 2. Add the package to your `nuxt.config.ts`:
     ```ts
-      modules: ['@sidebase/nuxt-auth'],
-    })
+        modules: ["@sidebase/nuxt-auth"],
```
BracketJohn, line 21: why did you decide to reformat the whole readme? Please revert this

src/module.ts
```diff
@@ -59,7 +59,13 @@ export default defineNuxtModule<ModuleOptions>({
     const isOriginSet = typeof usedOrigin !== 'undefined'
     if (!isOriginSet) {
       // TODO: see if we can figure out localhost + port dynamically from the nuxt instance
-      usedOrigin = 'http://localhost:3000'
```
BracketJohn, line 62: lets:
- remove the todo as we finished
- not make the protocol dynamic -> this is always http in development

BracketJohn (changes requested):

README.md
IronicUsername: yeah this is auto enabled at my end..
reverting this.

README.md
```diff
@@ -685,7 +690,7 @@ Note: The playground has considerably less polishing than the example page.
 - Run `npm run dev:prepare` to generate type stubs.
 - Use `npm run dev` to start [the module playground](./playground) in development mode.
 - Run `npm run lint` to run eslint
-- Run `npm run type` to run typescheck via tsc
```
BracketJohn, line 688: great catch, thanks!

BracketJohn (approved):

BracketJohn:
One last check: Can you test this locally and ensure that it works? To do so, just:
1. start the app two times,
2. open both apps,
3. log into both apps with credentials, make sure that:
    - [ ] login works
    - [ ] both apps direct you to the right origin: One should be `localhost:3000`, one should be `localhost:3001`

IronicUsername:
>3. log into both apps with credentials, make sure that:
>- [x] login works
>- [x] both apps direct you to the right origin: One should be localhost:3000, one should be localhost:3001

**NOTE**: 
in case this is relevant, i could sign-in on `locahost:3000` and after just reloading `localhost:3001` i was signed in there as well.
seems to me like this is desired but just to check if this is also what you want.

BracketJohn:
>  seems to me like this is desired but just to check if this is also what you want.

I _think_ that this is because the same cookie is used by both sites -> can you check your network tab during the refresh to ensure that all requests go to `localhost:3001` in the second tab?

IronicUsername:
oh no it was bc the port on the `3001` was not set to `3001` but default `3000` 

is there maybe some way i'm unaware of to get the internal "updated" port from the application if it detects an app running on the same default port?

if that would be the case, we should set this as default value and not a hardcoded int...

BracketJohn:
> is there maybe some way i'm unaware of to get the internal "updated" port from the application if it detects an app running on the same default port?

the goal of this pr is to use the internal mechanism you added to determione the port. So that means that `nuxt.options.server.port` is not working?

IronicUsername:
>So that means that nuxt.options.server.port is not working?

unsure sir...
so 1. no `nuxt.options.server.port` does not change when an app already blocks the port.
2. according to the nuxt code, `nuxt.options.server.port` and `nuxt.options.server.host` are both deprecated... 
what makes this weirder is that in this [thread](https://github.com/nuxt/framework/issues/3572) they imply that its basically not deprecated...

To me it seems like we have to get the `port` from somewhere else if we want to have this "automated workflow" for blocking ports...

maybe nuxt somewhere exposes the "newly" created port somewhere...

BracketJohn:
can you find a spot? somewhere nuxt has to run the logic that determines whether port 3000 is taken and then assign another port

IronicUsername:
sadly not yet..
but we could replicate the logic at our end.

the port selection is being done with a lib called [`get-port`](https://www.npmjs.com/package/get-port). 
we could have a function that "tests" out every port starting at port `N` and will:
- try port `N`
- if port is already in use
  - count up `N+1`
  - restart
- if port not in use
  - return it

The down side here is, its never the actual true port since we generate it ourselves..
The up side, its ez pz and will probably work for most cases.

IronicUsername:
ok found out that `get-port` has already smth similar documented:
https://github.com/sindresorhus/get-port/blob/0760c987c17581395d4e30432881dcb0ca6ca94a/index.d.ts#L54-L62

<details>
<summary>went with this approach but it has 1 big-ish edgecase:</summary>

<description>

Lets say we do the following:
- Open 4 different terminal windows
- start from each terminal one by one `npm run dev`
- we now occupy ports `3000` - `3003`

`getPort` would now calculate
- `3001` for the first window (but actually runs on `3000`)
- `3002` for the second window (but actually runs on `3001`)
- etc...

meaning we are always "behind one" because as it looks like `get-port` itself starts a server to check if a port is free. [[sauce](https://github.com/sindresorhus/get-port/blob/0760c987c17581395d4e30432881dcb0ca6ca94a/index.js#L44), [sauce](https://nodejs.org/api/net.html#netcreateserveroptions-connectionlistener)]

this is not a huge problem since we can then just "assume" the port we get from `get-port` is actually -1.

The problem starts when you now `Ctrl-C` the service running on `3000` and `3002` and  restart on the window where the `3002` was running.

2022-11-24 at 4 46 08 PM

Exhibit A. Everything normal as it should be. 

now i stop 1 and 3...
... and restart 3 before 1

2022-11-24 at 4 49 44 PM
Exhibit B. Everything is not as it should be

</description>
</details>

BracketJohn:
checkout: https://github.com/nuxt/framework/blame/265db5051510dcf8abf5d5c4563f0f68baf7c23b/packages/schema/src/config/dev.ts

can we make use of this? (:

IronicUsername:
>checkout: https://github.com/nuxt/framework/blame/265db5051510dcf8abf5d5c4563f0f68baf7c23b/packages/schema/src/config/dev.ts
> can we make use of this? (: 

nope.. stays `undefined`...

> Maybe we need to upgrade nuxt-auth to stable nuxt though

yes, tried this as well... 
Upgraded `nuxt` to `3.0.0` and funny enough `nuxt.options.server.host` and `nuxt.options.server.port` really are deprecated as mentioned [above](https://github.com/sidebase/nuxt-auth/pull/51#issuecomment-1326375515)...

kinda running out of options.
i think it would be best to identify where this port selection even happens...

maybe you got some pointers? @BracketJohn

BracketJohn:
Yeah, it could be that the port is only defined at runtime (where the nuxtauthhandler is evaluated) and not at build time (where the module.ts code runs).

So to quickly verify this: Can you try to access the above options inside the `nuxtAuthHandler.ts` code? just do some console.logs to quickly verify whether they are defined at that point

IronicUsername:
imported like this:
```js
// nuxtAuthHandler.ts
import defineUntypedSchema from 'nuxt'
// [...]
console.log('defineUntypedSchema', defineUntypedSchema)
// [...]
```

```sh
# cli output
Nuxi 3.0.0                                                                                                                                                                                                                                           13:03:50
Nuxt 3.0.0 with Nitro 1.0.0                                                                                                                                                                                                                          13:03:50
                                                                                                                                                                                                                                                     13:03:51
  > Local:    http://localhost:3000/
  > Network:  http://10.195.62.250:3000/

ℹ Setting up auth...                                                                                                                                                                                                                       nuxt-auth 13:03:51
ℹ Using /api/auth as the auth API location, make sure the [...].ts file with the export default NuxtAuthHandler({ ... }) is added there. Use the nuxt.config.ts auth.origin and auth.basePath config keys to change the API location       nuxt-auth 13:03:51
✔ Auth module setup done                                                                                                                                                                                                                   nuxt-auth 13:03:51
ℹ Vite client warmed up in 405ms                                                                                                                                                                                                                     13:03:52
✔ Nitro built in 273 ms                                                                                                                                                                                                                        nitro 13:03:52

 ERROR  [worker reload] [worker init] The requested module 'file:///Users/IronicUsername/Development/personal/nuxt-auth/node_modules/nuxt/dist/index.mjs' does not provide an export named 'default'                                                          13:03:52

  import defineUntypedSchema from 'node_modules/nuxt/dist/index.mjs';
  ^^^^^^^^^^^^^^^^^^^
  SyntaxError: The requested module 'node_modules/nuxt/dist/index.mjs' does not provide an export named 'default'
  at ModuleJob._instantiate (node:internal/modules/esm/module_job:127:21)
  at async ModuleJob.run (node:internal/modules/esm/module_job:193:5)
  at async Promise.all (index 0)
  at async ESMLoader.import (node:internal/modules/esm/loader:337:24)
  at async loadESM (node:internal/process/esm_loader:88:5)
  at async handleMainPromise (node:internal/modules/run_main:61:12)
```

BracketJohn:
closing as mostly outdated. We already automatically infer host + port for the app-side useSession, only thing missing is the backend `NuxtAuthHandler`. Let's resolve it in a new PR!

PR #60, opened 2022-11-25
Added documentation and sync workflows

zoey-kaiser:


zoey-kaiser:
The workflow is currently failing, because it expects the source branch to be the main branch. As it is not, the action is attempting to also create a new PR on `nuxt-auth` in order to sync both main branches of both repos. This issue will not longer exist when we merge into main and have it push the changes over then, as the main branch of `nuxt-auth` will already be up to date!

PR #75, opened 2022-12-02
fix: don't set `loading` to `true` on use `getSession`

### LINKED ISSUE #74: Status should only be `loading` if the session was not fetched yet
### Environment

_No response_

### Reproduction

Use `getSession`.

### Describe the bug

Every time we make a request to get the session, the status is changed to `loading`. This should not happen for two reasons:

1. This behaviour doesn't match the `status` of NextAuth.js. The status is only set to `loading` when the session was not fetched yet.
2. When we get the session after being authenticated, the session is already "loaded", we only want to be sure that data is up-to-date. So if we perfom a check to ensure that `status` is `authenticated`, changing it to `loading` would cause this check to fail.

### Additional context

_No response_

### Logs

_No response_

JoaoPedroAS51:
See #74 for context.

Closes #74  .

Checklist:

### DIFF
--- a/src/runtime/composables/useSession.ts
+++ b/src/runtime/composables/useSession.ts
@@ -185,7 +185,6 @@ const getSession = async (getSessionOptions?: GetSessionOptions) => {
     },
     onRequest: ({ options }) => {
       lastRefreshedAt.value = new Date()
-      loading.value = true
 
       options.params = {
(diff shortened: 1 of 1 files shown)


BracketJohn (approved):
LGTM ❤️

PR #76, opened 2022-12-04
feat: add `trustHost` option and fallback to `host` if not in production

### LINKED ISSUE #50: Use nuxt nitro config to set host and port in local development

### LINKED ISSUE #72: Investigate whether we can drop the requirement of setting an `origin` for production
### Describe the feature

At the moment we require an origin for production and fallback to defaults for development.

This was based on the deployment guide of next auth: https://next-auth.js.org/deployment

In a recent PR @JoaoPedroAS51 raised that we may be able to drop it: https://github.com/sidebase/nuxt-auth/pull/69#discussion_r1037297982

I found no convincing reasons to not drop it, however this should be investigated first as it may be security related (e.g.: for oauth callbacks).

### Additional information

_No response_

JoaoPedroAS51:
`origin` is required for production, but optional for development. If `origin` is not provided, it will fallback to `x-forwarded-host` or `host` headers, which solves the host/port issue.

`trustHost` can be enabled to use either the `x-forwarded-host` or `host` headers, instead of `origin`

Closes #50, closes #72

Checklist:

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -72,7 +72,7 @@ const defaults: ModuleOptions & { basePath: string } = {
   isEnabled: true,
   origin: undefined,
   basePath: '/api/auth',
-  trustHost: Boolean(process.env.AUTH_TRUST_HOST ?? process.env.VERCEL),
+  trustHost: false,
   enableSessionRefreshPeriodically: false,
   enableSessionRefreshOnWindowFocus: true,
   enableGlobalAppMiddleware: false
--- a/src/module.ts
+++ b/src/module.ts
@@ -153,17 +153,6 @@ export default defineNuxtModule<ModuleOptions>({
     // 6. Add plugin for initial load
     addPlugin(resolve('./runtime/plugin'))
 
-    // 7. Setup next-auth env-variables just to be safe, see https://github.com/nextauthjs/next-auth/blob/6280fe9e10bd123aeab576398d1e807a5ac37edc/apps/playground-nuxt/src/module.ts#L32-L38
-    if (!nuxt.options.dev && isOriginSet) {
-      extendViteConfig((config) => {
-        config.define = config.define || {}
-        config.define['process.env.NEXTAUTH_URL'] = JSON.stringify(url)
-        config.define['process.env.NEXTAUTH_URL_INTERNAL'] = JSON.stringify(url)
-        config.define['process.env.AUTH_TRUST_HOST'] = JSON.stringify(options.trustHost)
-        config.define['process.env.VERCEL_URL'] = JSON.stringify(process.env.VERCEL_URL)
-      })
-    }
-
     logger.success('`nuxt-auth` setup done')
   }
 })
--- a/src/module.ts
+++ b/src/module.ts
@@ -1,4 +1,4 @@
-import { defineNuxtModule, useLogger, addImportsDir, createResolver, addTemplate, addPlugin, extendViteConfig } from '@nuxt/kit'
+import { defineNuxtModule, useLogger, addImportsDir, createResolver, addTemplate, addPlugin } from '@nuxt/kit'
 import defu from 'defu'
 import { joinURL } from 'ufo'
 
--- a/src/module.ts
+++ b/src/module.ts
@@ -30,6 +30,16 @@ interface ModuleOptions {
    */
   basePath: string | undefined
   /**
+   * If set to `true`, `NuxtAuth` will use either the `x-forwarded-host` or `host` headers,
+   * instead of `auth.origin`
+   * Make sure that reading `x-forwarded-host` on your hosting platform can be trusted.
+   * - ⚠ **This is an advanced option.** Advanced options are passed the same way as basic options,
+   * but **may have complex implications** or side effects.
+   * You should **try to avoid using advanced options** unless you are very comfortable using them.
+   * @default Boolean(process.env.AUTH_TRUST_HOST ?? process.env.VERCEL)
+   */
+  trustHost: boolean
+  /**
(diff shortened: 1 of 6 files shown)


src/module.ts
```diff
@@ -62,6 +72,7 @@ const defaults: ModuleOptions & { basePath: string } = {
   isEnabled: true,
   origin: undefined,
   basePath: '/api/auth',
+  trustHost: Boolean(process.env.AUTH_TRUST_HOST ?? process.env.VERCEL),
```
BracketJohn, line 75: I think this will not work as expected: This is only evaluated at "build time" of the module AFAIK. So then when time for deployment rolls around, `trustHost` option is _not_ going to be configurable via `process.env.AUTH_TRUST_HOST` any longer.

Instead, we'll have to read this at runtime, AFAIU we can:
- process this at runtime 
- add it to the runtimeConfig, so that it is picked up by nuxt from the env on startup (see https://nuxt.com/docs/migration/runtime-config#runtime-config)

I think the second option is better.

This topic in general imho is a bit messy in nuxt 3 with how modules handle config + runtime config + env vars.

src/module.ts
```diff
@@ -148,12 +154,15 @@ export default defineNuxtModule<ModuleOptions>({
...
-      config.define = config.define || {}
-      config.define['process.env.NEXTAUTH_URL'] = JSON.stringify(url)
-      config.define['process.env.NEXTAUTH_URL_INTERNAL'] = JSON.stringify(url)
-      config.define['process.env.VERCEL_URL'] = JSON.stringify(process.env.VERCEL_URL)
-    })
+    if (!nuxt.options.dev && isOriginSet) {
+      extendViteConfig((config) => {
+        config.define = config.define || {}
+        config.define['process.env.NEXTAUTH_URL'] = JSON.stringify(url)
+        config.define['process.env.NEXTAUTH_URL_INTERNAL'] = JSON.stringify(url)
```
BracketJohn, line 161: out of interest: where you able to verify that `NextAuth` correctly picks these up? When I took over this code from the app-example I did not attempt to verify it's supposed functionality.

BracketJohn (changes requested):
Thanks! Just some small stuff.

I also tested:
- npm run dev
- npm run build:playground && npm run start:playground

To verify that the module still works as a whole. When trying:
```
> npm run build:playground && npm run start:playground
```

and then directly visiting http://0.0.0.0:3000/ I get an error:
```
[nuxt] [request error] [unhandled] [500] nuxt instance unavailable
  at useNuxtApp (./playground/.output/server/chunks/app/server.mjs:157:13)
  at useRequestEvent (./playground/.output/server/chunks/app/server.mjs:420:36)
  at getRequestURL (./playground/.output/server/chunks/app/server.mjs:982:25)
  at Object.onRequest (./playground/.output/server/chunks/app/server.mjs:1094:37)
  at $fetchRaw2 (./playground/.output/server/node_modules/ofetch/dist/shared/ofetch.5cb01515.mjs:122:29)
  at onError (./playground/.output/server/node_modules/ofetch/dist/shared/ofetch.5cb01515.mjs:102:16)
  at ./playground/.output/server/node_modules/ofetch/dist/shared/ofetch.5cb01515.mjs:147:14
  at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

This looks to be the nested composable calls problem. The problem also occurs on `main` 😢

src/module.ts
JoaoPedroAS51: Yeah, me neither... I don't know if it actually works...

JoaoPedroAS51: I searched for envs in NextAuth repo and it seems that the core is not using envs. So I think it's safe to remove them.

src/module.ts
BracketJohn: Yeah, let's add the env for now and see if there's a later request for env-variable support and then work on it.

src/module.ts
BracketJohn: What about: https://github.com/nextauthjs/next-auth/blob/7e91d7df54f9d8ca782c93a63f583ce89c2d178d/packages/next-auth/src/utils/node.ts#L37-L39

References:
image

-> If based on this it remains unclear to us, I would leave this part as:
- it contains only information that can be public
- does not incur added runtime cost
- is also part of the nextauth nuxt example

BracketJohn:
I merged in main with the fix to the nested compsable bug from https://github.com/sidebase/nuxt-auth/pull/76#pullrequestreview-1204543610. It is fixed 🎊

src/module.ts
JoaoPedroAS51: From what I see, this util is only used for `next` module. And as the next-auth wants to be agnostic, I believe they will not be using envs in core.

BracketJohn: :+1: very good, thanks for further looking into  it. Then let's drop it (:

BracketJohn (approved):

PR #87, opened 2022-12-07
fix(middleware): redirect to signIn using `location.replace`

### LINKED ISSUE #84: Look into pages not being correctly protected
### Environment

_No response_

### Reproduction

_No response_

### Describe the bug

A user on our discord reported the following bug in the newest version of `nuxt-auth` v0.3.0 Alpha 3. We should look into resolving this before our release:

https://a.cl.ly/7KujyX7e?method=GET

### Additional context

_No response_

### Logs

_No response_

JoaoPedroAS51:
This will navigate to signIn page without adding the protected route to the history. See #84 for context.

This PR improves `navigateTo` util:
- add `replace` option
- replicate the original `navigateTo` behaviour by using vue router if `external` is `false`
  - (we are not using this at all, but I thought would be good to have it just in case)
- return `Promise` in client side

Closes #84 

Checklist:


JoaoPedroAS51:
This does not fix the real issue though... just tested here and can reproduce by:
signIn -> go to protected page -> signOut -> go back two times (<-)

JoaoPedroAS51:
But still, this is an improvement as it prevents the protected route from being added to the history, replicating server side behaviour

src/runtime/composables/useSession.ts
```diff
@@ -33,6 +33,8 @@ interface SignInOptions extends Record<string, unknown> {
   callbackUrl?: string
   /** [Documentation](https://next-auth.js.org/getting-started/client#using-the-redirect-false-option) */
   redirect?: boolean
+
+  replace?: boolean
```
BracketJohn, line 37: I propose to either:
- not make this an option for the lib-user (downsides are unclear naming + we'd have to replicate the behavior for `useSession` as well, as it has the same options as `signIn`)
- also add this to `useSession` options + behavior



I slightly prefer option 1 (:

JoaoPedroAS51: I would go with option 2, as we should not use `replace` by default when the user calls signIn(), only in auth middleware. And if the user want to create a custom auth middleware, he can enable it.

src/runtime/composables/useSession.ts
```diff
@@ -186,7 +200,6 @@ const getSession = (getSessionOptions?: GetSessionOptions) => {
 
       options.params = {
         ...(options.params || {}),
-        // The request is executed with `server: false`, so window will always be defined at this point
```
BracketJohn, line 189: ❤️

BracketJohn (approved):

PR #92, opened 2022-12-08
docs: Fixed redirect link in docs

zoey-kaiser:


BracketJohn (approved):
any other links / place we should check?

zoey-kaiser:
> any other links / place we should check?

Not as I know! I tested all of the links we can see, I just failed to check the redirect from `/nuxt-auth`

PR #115, opened 2022-12-19
fix: use type `Session` from `next-auth`

### LINKED ISSUE #112: Streamline typing across next auth session data and `useSession` data type
### Describe the feature

Raised by @noook in [#61:](https://github.com/sidebase/nuxt-auth/issues/61#issuecomment-1356826150)

> The signature for `useSession` does not go alone this answer :(

image

> `useSession().data` does not refer to next-auth types so overriding it does not help much when using the type outside of the NuxtAuth handler

We should:
- [ ] ensure that both have the same data-type, so that it can be overwritten easily.
- [ ] update the recipe on how to add custom session data https://sidebase.io/nuxt-auth/recipes/custom-session-data
  - source is here: https://github.com/sidebase/nuxt-auth/blob/main/docs/content/5.recipes/4.custom-session-data.md

### Additional information

_No response_

JoaoPedroAS51:
Add types `UseSessionReturn` and `UseSessionStateReturn`

Closes #112 

Checklist:

### DIFF
--- a/src/runtime/composables/useSession.ts
+++ b/src/runtime/composables/useSession.ts
@@ -2,10 +2,16 @@ import type { AppProvider, BuiltInProviderType } from 'next-auth/providers'
 import defu from 'defu'
 import { callWithNuxt } from '#app'
 import { readonly } from 'vue'
+import type { ComputedRef, Ref } from 'vue'
 import { navigateTo, getRequestURL, joinPathToApiURL } from '../utils/url'
 import { _fetch } from '../utils/fetch'
 import { isNonEmptyObject } from '../utils/checkSessionResult'
-import useSessionState, { SessionData } from './useSessionState'
+import useSessionState from './useSessionState'
+import type {
+  SessionData,
+  SessionLastRefreshedAt,
+  SessionStatus
+} from './useSessionState'
 import { createError, useRequestHeaders, useNuxtApp } from '#imports'
 
 /**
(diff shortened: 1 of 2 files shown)


src/runtime/composables/useSession.ts
```diff
@@ -69,7 +75,8 @@ const getCsrfToken = () => {
 }
 
 /**
- * Trigger a sign in flow for the passed `provider`. If no provider is given the sign in page for all providers will be shown.
+ * Trigger a sign in flow for the passed `provider`. If no provider is given the sign in page for all providers will be
+ * shown.
```
BracketJohn, line 79: nit: was this moved to a newline on accident? :D

BracketJohn (commented):
So after this PR types will correctly reference what we setup in our own module, but will the approach to overwrite them by overwriting the `next-auth` types outlined here work afterwards: https://sidebase.io/nuxt-auth/recipes/custom-session-data? 🤔

src/runtime/composables/useSession.ts
JoaoPedroAS51: > So after this PR types will correctly reference what we setup in our own module, but will the approach to overwrite them by overwriting the `next-auth` types outlined here work afterwards: https://sidebase.io/nuxt-auth/recipes/custom-session-data? 🤔

Yes, it will work by overwriting the `next-auth` types. 

A little bit of context: My guess is that nuxt auto-imports was generating the result of `Session` type instead of referencing it. So I fixed it to use the `Session` from `next-auth`.

BracketJohn (approved):

BracketJohn:
Thanks for the explanation (: Change seems good, I also checked it locally now (:

Issue #128, opened 2022-12-24
Nuxt Auth does not work with Prisma Adapter


Q16solver:
### Environment

Operating System: `MacOs`
Node Version: `v18.9.1`
Nuxt Version: `3.0.0`
Nitro Version: `1.0.0`
Package Manager: `npm@8.19.1`
Builder: `vite`
User Config: `css`, `build`, `modules`, `runtimeConfig`, `auth`
Runtime Modules: `@sidebase/nuxt-auth@0.3.2`

### Describe the bug

If I install the Prisma adapter package, then I believe there's some weird conflict that exists with both `prisma adapter` & `nuxt-auth` package both having `next-auth` as a dependency. Perhaps there could be some way to resolve it from this package side by making next-auth a peer dependency and requiring users to install it or something similar?

![image](https://user-images.githubusercontent.com/50786287/210480366-3947cb36-cee4-4dbc-a3a5-b8fc3ece94c8.png)

BracketJohn:
Hey @Q16solver 👋 

Thanks for the issue - will likely make next auth a peer dep to resolve this soon (: 

Today is off for Christmas though (: Merry christmas to you (:

Q16solver:
Yep, thanks for the fast response even though it’s Christmas! Merry Christmas to to you too haha

BracketJohn:
Thanks ❤️

MuhammadKhizar7:
I am using a Prisma adapter with nuxt-auth, and it works fine for me. I did not yet go to deploy it, and I have example to show. 
I use Windows and use NPM it might be mac issue.
https://github.com/MuhammadKhizar7/nuxt-prisma

Q16solver:
haha well, actually you walked into one of the "workarounds" which is installing nuxt-auth first, so the package-lock.json prioritises `nuxt-auth`'s dependency first for whatever reason, and installing the prisma adapter package next doesn't affect it. You'll find that if you remove your node_modules and package-lock.json and regenerate everything, it won't work

BracketJohn:
Hey you two 🔢 

#179 is now open which makes `next-auth` a peer dependency (:

PR #136, opened 2023-01-03
fix: do not use async composables in auth middleware

### LINKED ISSUE #132: `Error: nuxt instance unavailable` when signing out from a protected page
### Environment

- Operating System: `Darwin`
- Node Version:     `v16.13.2`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `npm@9.2.0`
- Builder:          `vite`
- User Config:      `css`, `modules`, `auth`
- Runtime Modules:  `@nuxtjs/tailwindcss@6.1.0`, `@sidebase/nuxt-auth@0.3.2`

### Reproduction

https://github.com/LLCampos/nuxt-auth-example

This [commit](https://github.com/LLCampos/nuxt-auth-example/commit/fd1d0118e07127e3a7fd51c1269cc6eb08e6e70a) contains the changes necessary to the current example app to reproduce the issue.

### Describe the bug

* Start application
* It should immediately take you to login page
* Login with credentials
* Logout

I get the following in logs

```
[Vue Router warn]: uncaught error during route navigation:                                                                                                            11:09:23
Error: nuxt instance unavailable                                                                                                                                      11:09:23
    at useNuxtApp (/Users/luiscampos/Development/nuxt-auth-example/node_modules/nuxt/dist/app/nuxt.mjs:165:13)
    at Module.useRuntimeConfig (/Users/luiscampos/Development/nuxt-auth-example/node_modules/nuxt/dist/app/nuxt.mjs:173:10)
    at getApiURL (/Users/luiscampos/Development/nuxt-auth-example/node_modules/@sidebase/nuxt-auth/dist/runtime/utils/url.mjs:10:40)
    at Module.joinPathToApiURL (/Users/luiscampos/Development/nuxt-auth-example/node_modules/@sidebase/nuxt-auth/dist/runtime/utils/url.mjs:15:66)
    at /Users/luiscampos/Development/nuxt-auth-example/node_modules/@sidebase/nuxt-auth/dist/runtime/middleware/auth.mjs:16:65
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Object.callAsync (file:///Users/luiscampos/Development/nuxt-auth-example/node_modules/unctx/dist/index.mjs:53:16)
    at async /Users/luiscampos/Development/nuxt-auth-example/node_modules/nuxt/dist/pages/runtime/router.mjs:110:22
[Vue Router warn]: uncaught error during route navigation:                                                                                                            11:09:23
Error: nuxt instance unavailable                                                                                                                                      11:09:23
    at useNuxtApp (/Users/luiscampos/Development/nuxt-auth-example/node_modules/nuxt/dist/app/nuxt.mjs:165:13)
    at Module.useRuntimeConfig (/Users/luiscampos/Development/nuxt-auth-example/node_modules/nuxt/dist/app/nuxt.mjs:173:10)
    at getApiURL (/Users/luiscampos/Development/nuxt-auth-example/node_modules/@sidebase/nuxt-auth/dist/runtime/utils/url.mjs:10:40)
    at Module.joinPathToApiURL (/Users/luiscampos/Development/nuxt-auth-example/node_modules/@sidebase/nuxt-auth/dist/runtime/utils/url.mjs:15:66)
    at /Users/luiscampos/Development/nuxt-auth-example/node_modules/@sidebase/nuxt-auth/dist/runtime/middleware/auth.mjs:16:65
    at async Object.callAsync (file:///Users/luiscampos/Development/nuxt-auth-example/node_modules/unctx/dist/index.mjs:53:16)
    at async /Users/luiscampos/Development/nuxt-auth-example/node_modules/nuxt/dist/pages/runtime/router.mjs:110:22
[nitro] [dev] [unhandledRejection] Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client                                                11:09:23
    at new NodeError (node:internal/errors:371:5)
    at ServerResponse.setHeader (node:_http_outgoing:576:11)
    at setResponseHeader (file:///Users/luiscampos/Development/nuxt-auth-example/node_modules/nitropack/node_modules/h3/dist/index.mjs:277:18)
    at Object.errorhandler [as onError] (file:///Users/luiscampos/Development/nuxt-auth-example/.nuxt/dev/index.mjs:450:5)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Server.toNodeHandle (file:///Users/luiscampos/Development/nuxt-auth-example/node_modules/nitropack/node_modules/h3/dist/index.mjs:752:9) {
  code: 'ERR_HTTP_HEADERS_SENT'
```

### Additional context

_No response_

### Logs

_No response_

BracketJohn:
Closes #132

This PR:
- manually assembles a navigate to call, taking into account:
  - the original target of the request for `callback`
  - setting the correct `error` code for the sign-in page that is redirected to
- adds early 404 redirects to not enforce authentication for 404 pages per default
- adds a configuration option to change the aforementioned 404-behavior

This fully resolves the `ERR_HTTP_HEADERS_SENT` problem 🎊 It seems to be related to redirecting to a 404-page and may be the same as https://github.com/nuxt/framework/issues/9438. The problem will still occur when enforcing authentication for 404 pages by setting `allow404WithoutAuth: false`.

Checklist:

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -14,7 +14,7 @@ interface GlobalMiddlewareOptions {
    * @example false
    * @default true
    */
-  enableEarly404Redirect?: boolean
+  allow404WithoutAuth?: boolean
 }
 
 interface ModuleOptions {
@@ -96,7 +96,7 @@ const defaults: ModuleOptions & { basePath: string } = {
   enableSessionRefreshOnWindowFocus: true,
   enableGlobalAppMiddleware: false,
   globalMiddlewareOptions: {
-    enableEarly404Redirect: true
+    allow404WithoutAuth: true
   }
 }
 export default defineNuxtModule<ModuleOptions>({
--- a/src/module.ts
+++ b/src/module.ts
@@ -2,6 +2,21 @@ import { defineNuxtModule, useLogger, addImportsDir, createResolver, addTemplate
 import defu from 'defu'
 import { joinURL } from 'ufo'
 
+interface GlobalMiddlewareOptions {
+  /**
+   * Whether to enforce authentication if the target-route does not exist. Per default the middleware redirects
+   * to Nuxts' default 404 page instead of forcing a sign-in if the target does not exist. This is to avoid a
+   * user-experience and developer-experience of having to sign-in only to see a 404 page afterwards.
+   *
+   * Note: Setting this to `false` this may lead to `vue-router` + node related warnings like: "Error [ERR_HTTP_HEADERS_SENT] ...",
+   * this may be related to https://github.com/nuxt/framework/issues/9438.
+   *
+   * @example false
+   * @default true
+   */
+  enableEarly404Redirect?: boolean
+}
+
 interface ModuleOptions {
   /**
    * Whether the module is enabled at all
@@ -65,6 +80,10 @@ interface ModuleOptions {
    * @default false
    */
   enableGlobalAppMiddleware: boolean
+  /**
+   * Options of the global middleware. They will only apply if `enableGlobalAppMiddleware` is set to `true`.
+   */
+  globalMiddlewareOptions: GlobalMiddlewareOptions
 }
 
 const PACKAGE_NAME = 'nuxt-auth'
@@ -75,7 +94,10 @@ const defaults: ModuleOptions & { basePath: string } = {
   trustHost: false,
   enableSessionRefreshPeriodically: false,
   enableSessionRefreshOnWindowFocus: true,
-  enableGlobalAppMiddleware: false
+  enableGlobalAppMiddleware: false,
+  globalMiddlewareOptions: {
+    enableEarly404Redirect: true
+  }
 }
 export default defineNuxtModule<ModuleOptions>({
   meta: {
--- a/src/runtime/middleware/auth.ts
+++ b/src/runtime/middleware/auth.ts
@@ -1,7 +1,6 @@
 import { defineNuxtRouteMiddleware, navigateTo } from '#app'
 import { withQuery } from 'ufo'
 import useSession from '../composables/useSession'
-import { joinPathToApiURL } from '../utils/url'
 
 export default defineNuxtRouteMiddleware((to) => {
   if (to.meta.auth === false) {
@@ -14,12 +13,15 @@ export default defineNuxtRouteMiddleware((to) => {
   }
 
   /**
-   * Nuxt 3 default behavior is to continue with the route navigation if the middleware does not return a `navigateTo` (see https://nuxt.com/docs/guide/directory-structure/middleware#format).
+   * We cannot directly call and/or return `signIn` here as:
+   * - `signIn` uses async composables under the hood, leading to "nuxt instance undefined errors", see https://github.com/nuxt/framework/issues/5740#issuecomment-1229197529
+   * - if something different than `navigateTo` is returned the navigation will not be blocked, so a content-flash would occur, see https://nuxt.com/docs/guide/directory-structure/middleware#format
    *
-   * For this reason we build and return a custom `navigateTo` instead of returning `signIn(...)`
-   * */
-  const url = withQuery(joinPathToApiURL('signin'), {
-    callbackUrl: to.path
+   * So we assemble our own call and return `navigateTo`.
+   *  */
+  const url = withQuery('/api/auth/signin', {
+    callbackUrl: to.path,
+    error: 'SessionRequired'
   })
-  return navigateTo('/api/auth/signin')
+  return navigateTo(url)
 })
--- a/src/runtime/middleware/auth.ts
+++ b/src/runtime/middleware/auth.ts
@@ -22,14 +22,14 @@ export default defineNuxtRouteMiddleware((to) => {
   const authConfig = useRuntimeConfig().public.auth
 
   /**
-   * We do not want to enforce protection on `404` pages (unless the user opts out of it by setting `enableEarly404Redirect: false`).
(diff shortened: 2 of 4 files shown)


src/module.ts
```diff
@@ -75,7 +94,10 @@ const defaults: ModuleOptions & { basePath: string } = {
   trustHost: false,
   enableSessionRefreshPeriodically: false,
   enableSessionRefreshOnWindowFocus: true,
-  enableGlobalAppMiddleware: false
+  enableGlobalAppMiddleware: false,
```
JoaoPedroAS51, line 97: Maybe we can merge `enableGlobalAppMiddleware` with `globalMiddlewareOptions`, so we have all in one place. What do you think?

Something like this:
```ts
globalAppMiddleware: {
  enabled: false,
  allow404WithoutAuth: true
}
```

src/runtime/middleware/auth.ts
```diff
@@ -1,32 +1,68 @@
...
+  // adapted from: https://github.com/nuxt/framework/blob/ab2456c295fc8c7609a7ef7ca1e47def5d087e87/packages/nuxt/src/app/composables/router.ts#L97-L115
+  if (process.server) {
+    const nuxtApp = useNuxtApp()
+    if (nuxtApp.ssrContext && nuxtApp.ssrContext.event) {
+      return nuxtApp.callHook('app:redirected').then(() => sendRedirect(nuxtApp.ssrContext!.event, url, 302))
+    }
+  }
+
+  window.location.href = url
+  return navigateTo(undefined)
```
JoaoPedroAS51, line 67: In my tests I noticed that `navigateTo(undefined)` will redirect the user to `/` before redirecting to signIn, in client side. `navigateTo` defaults to `/` when the path is not provided. See https://github.com/nuxt/framework/blob/ab2456c295fc8c7609a7ef7ca1e47def5d087e87/packages/nuxt/src/app/composables/router.ts#L77-L79

To reproduce: go to [page that is always unprotected](http://localhost:3000/always-unprotected), then click  [globally protected page](http://localhost:3000/protected/globally). You will see a content-flash of the home page, and if you go back it will go to home page.

I'm not sure how to solve this :/

JoaoPedroAS51 (changes requested):
Nice work! And good catch about the 404 page!

Just found one issue in client side redirect, I added a comment there

src/module.ts
BracketJohn: I like the idea very much, I think it could even have a signature of:
```
globalAppMiddleware: boolean | GlobalAppMiddlewareOptions
```

with `false` turning it off, `true` turning it on with default config and `{ ... }` allowing for granular configuration. 

An issue with this approach and why I didn't go with it so far:
- breaks backwards-compat (less severe if we go with the `boolean | ...` signature)
- maybe does not work with the way that `nuxt/schema` generates the typings for the config keys -> there's some really weird errors there

I would look into whether I can mitigate the two problems and add it here, otherwise I propose that we move it to the future, maybe also as part of a `0.4.x` jump in versioning (incoming anyways, due to requirement to make `next-auth ` a peer-dep which is definitely a breaking change, see #128 for example)

JoaoPedroAS51: I liked the signature you suggested! Agreed.

One thing about versioning, maybe we should warn users to fix the exact version they are using to avoid breaking changes until we release v1.0.0

src/runtime/middleware/auth.ts
BracketJohn: Looked into it -> figured out that there is a bit less magic than I though: As long as we return a `Promise` that does not resolve immeadiatly, the content of the protected page will not be loaded.

I now set it to 100ms: It should never actually be awaited but will block the navigation for up to `100ms` or until the navigation to the singin-page completes (which should always be faster than 100ms, I hope?).

- Does this resolve the issue?
- Do you think 100ms is enough or should we increase it?

JoaoPedroAS51: Yeah, this actually works!!! But I think 100ms is not enough for slow connections. If I set network to slow 3g in the browser, I can see the content-flash for a long time.

If we set the `setTimeout` to `Infinity`, it solves completely. Can we possible cause a side-effect by using `Infinity` here? I have not used it before haha

BracketJohn: Woop wooop 🎊 

Good catch - I wanted to avoid using `Infinity`, so as a (good I think?) compromise I adapted the behavior:
- wait for up to 60 seconds
- if that does not work: delegate control back to `vue-router` -> this will still avoid content-flashes. The only downside is that it will log a `404` error in the console because vue-router does not know about the signin page

src/module.ts
BracketJohn: We're already following the policy of only releasing:
- patch changes that are non-breaking.
- minor changes if something breaking happened 

Do you think that's enough to avoid what you described?

src/runtime/middleware/auth.ts
JoaoPedroAS51: Yeah, I think this is good. Tested here and seems to work fine!

JoaoPedroAS51 (approved):

src/module.ts
JoaoPedroAS51: Yeah, based on what you said on discord about the approach of using `0.x` for now, I guess so.
But I think we could suggest users to add `nuxt-auth` using `~0.x` instead of `^0.x` in `package.json`, so it will only update to future patch versions, without incrementing the minor version.

PR #140, opened 2023-01-04
feat: make origin configurable at runtime

BracketJohn:
This PR:
- makes the `origin` configurable at runtime via `AUTH_ORIGIN`
- adds documentation to explain:
  - how to set the origin at build time
  - how to setup your own environment variable to configure the `secret` at build time


src/runtime/server/services/nuxtAuthHandler.ts
```diff
@@ -18,6 +18,11 @@ let preparedAuthHandler: ReturnType<typeof eventHandler> | undefined
 let usedSecret: string | undefined
 const SUPPORTED_ACTIONS: NextAuthAction[] = ['providers', 'session', 'csrf', 'signin', 'signout', 'callback', 'verify-request', 'error', '_log']
 
+export const ERROR_MESSAGES = {
```
zoey-kaiser, line 21: One feedback I got from the streamer using `nuxt-auth` for the first time was that our error messages were pretty long and spammed the console. 

Do we maybe want to shorten them here and have a dedicated part of our docs list fixes for each error? Then we can link that page in the error message and go more into detail into what the issue is?

zoey-kaiser (changes requested):

src/runtime/server/services/nuxtAuthHandler.ts
BracketJohn: Super idea! I've added it (:

PR #164, opened 2023-01-16
Add `defaultProvider`

### LINKED ISSUE #154: Define Default Provider for GlobalAppMiddleware
### Ask your question

In case I enable the enableGlobalAppMiddleware Flag in the nuxt config file, is it somehow possible to automatically try to login with a specific provider? This would allow me to implement a SSO solution. At the moment the default login screen is shown and the user has to chose a provider (but in my case there is only 1 available)
`
// nuxt.config.ts
auth: {
   enableGlobalAppMiddleware: true
}
`

### Additional information

I am using a custom OIDC provider

IronicUsername:
Closes #154

### Description
- add `defaultProvider`
- refactor `signIn` so that it is usable from application-side middleware's
- use `signIn` in application side middleware

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -81,6 +81,13 @@ interface ModuleOptions {
    */
   enableGlobalAppMiddleware: boolean
   /**
+   * The default login provider.
+   *
+   * @example true
+   * @default "DEFAULT_PROVIDER_NAME"
+   */
+  defaultProvider: string | undefined
+  /**
    * Options of the global middleware. They will only apply if `enableGlobalAppMiddleware` is set to `true`.
    */
   globalMiddlewareOptions: GlobalMiddlewareOptions
@@ -95,6 +102,7 @@ const defaults: ModuleOptions & { basePath: string } = {
   enableSessionRefreshPeriodically: false,
   enableSessionRefreshOnWindowFocus: true,
   enableGlobalAppMiddleware: false,
+  defaultProvider: undefined,
   globalMiddlewareOptions: {
     allow404WithoutAuth: true
   }
--- a/src/module.ts
+++ b/src/module.ts
@@ -81,10 +81,10 @@ interface ModuleOptions {
    */
   enableGlobalAppMiddleware: boolean
   /**
-   * The default login provider.
+   * The default auth provider that handles authentication.
    *
-   * @example true
-   * @default "DEFAULT_PROVIDER_NAME"
+   * @example "DEFAULT_PROVIDER_NAME"
+   * @default undefined
    */
   defaultProvider: string | undefined
   /**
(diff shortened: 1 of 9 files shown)


src/runtime/composables/useSession.ts
```diff
@@ -87,7 +87,13 @@ const signIn = async (
     return navigateToWithNuxt(errorUrl)
   }
 
-  // 2. Redirect to the general sign-in page with all providers in case either no provider or no valid provider was selected
+  // 2. Check if a custom provider was given and use its defaultProvider
+  const runtimeConfig = publicRuntimeConfig()
+  if (runtimeConfig?.auth?.defaultProvider !== undefined) {
+    provider = runtimeConfig.auth.defaultProvider
```
BracketJohn, line 93: we should never override the `provider` that was passed to us, we should only override it if `prover = undefined`  was passed

src/runtime/utils/url.ts
```diff
@@ -10,6 +10,7 @@ const getApiURL = () => {
 
 export const getRequestURL = (includePath = true) => _getURL(useRequestEvent()?.node.req, includePath)
 export const joinPathToApiURL = (path: string) => joinURL(getApiURL(), path)
+export const publicRuntimeConfig = () => useRuntimeConfig().public
```
BracketJohn, line 13: please remove this and directly access it instead:
- there's some caveats with using nested composables in Nuxt 3 that we want to avoid,
- we'd need to change more places that access the runtime config to make use of this

src/module.ts
```diff
@@ -80,6 +80,13 @@ interface ModuleOptions {
    * @default false
    */
   enableGlobalAppMiddleware: boolean
+  /**
+   * The default login provider.
+   *
+   * @example true
```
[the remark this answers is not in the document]
IronicUsername: missed the line^^

src/runtime/utils/url.ts
IronicUsername: the reason for me to put this here was the comment here
https://github.com/sidebase/nuxt-auth/blob/d303114130cb8b84d3f7afe5876d593939dd64c3/src/runtime/composables/useSession.ts#L78




... that i should just use the `useRuntimeConfig()` within the `useSession.ts` ?

if that is the case shouldn't we then also rework the lines that have the workaround for this?

zoey-kaiser (dismissed):

zoey-kaiser:
The code LGTM! 

However, I think we should add a bit more to the docs to let people know what this feature does. It may be self explanatory, but a bit more would probably be good! However, ill leave this to @BracketJohn to decide!

zoey-kaiser (approved):
I like the changes to the docs! Should make it a bit easier to understand. Thanks for the changes ❤️

BracketJohn (approved):

BracketJohn (approved):

PR #171, opened 2023-01-19
Change cookie name

### LINKED ISSUE #167: Change default cookie names
### Describe the feature

The cookie names should be "nuxt-auth" instead of "next-auth".

Better yet, provide a mechanism for changing the names to anything you want.

### Additional information

_No response_

IronicUsername:
Closes #167

### Description
- rm default cookie name
- add cookie naming opitons

### DIFF
--- a/docs/content/2.configuration/3.nuxt-auth-handler.md
+++ b/docs/content/2.configuration/3.nuxt-auth-handler.md
@@ -135,7 +135,7 @@ Simply by setting the environment variable `$NUXT_AUTH_COOKIE_NAME`.
 If this is not set, `nuxt-auth` resorts to its package name which is at this moment set to `nuxt-auth`:
 ```ts
 // file: ~/server/services/nuxtAuthHandler.ts
-import { PACKAGE_NAME } from '../../../module'
+import { PACKAGE_NAME } from '~~/../src/module'
 [...]
 const DEFAULT_NUXT_AUTH_COOKIE_NAME = process.env?.NUXT_AUTH_COOKIE_NAME || PACKAGE_NAME
 [...]
--- a/docs/content/2.configuration/3.nuxt-auth-handler.md
+++ b/docs/content/2.configuration/3.nuxt-auth-handler.md
@@ -124,6 +124,54 @@ export default NuxtAuthHandler({
 ```
 ::
 
+
+### Cookies
+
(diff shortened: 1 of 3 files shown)


docs/content/2.configuration/3.nuxt-auth-handler.md
```diff
@@ -124,6 +124,54 @@ export default NuxtAuthHandler({
 ```
 ::
 
+
+### Cookies
+
+Since the `nuxtAuthHandler` uses `next-auth` in the background, it also creates cookies in order to make the auth mechanisms work.
+There is the option to set a custom name to the cookies if there is the desire or need to do so.
+
+Simply by setting the environment variable `$NUXT_AUTH_COOKIE_NAME`.
```
zoey-kaiser, line 133: Do you need to do this? As far as I can see in the example below, you can also hardcode it inside the `nuxtAuthHandler.ts`. 

Will is also work "out-of-the-box" by just setting the .env without any other modifications to the `nuxtAuthHandler`?

IronicUsername: > Do you need to do this? 

no need, but the feature requester @nathanchase requested it and i couldnt think of a reason why not ^^

>Will is also work "out-of-the-box" by just setting the .env without any other modifications to the nuxtAuthHandler?

yes. this is the intention.

<hr/>

idk if its that relevant but we could also link to the OG `next-auth` docs [for this](https://next-auth.js.org/configuration/options#cookies) here..

zoey-kaiser: Sounds good. - Then I might add something like "alternatively" to the docs that it is clear that you can also set it in the `nuxtAuthHandler`!

Linking to the NextAuth docs is always very good, as it gives more insight into the underlying logic behind our package!

BracketJohn:
Let's only update the name of the cookie for now and leave configurability for a later date!

docs/content/2.configuration/3.nuxt-auth-handler.md
nathanchase: My main concern was the line from the NextAuth docs that said:
_"You can specify one or more cookies with custom properties, but if you specify custom options for a cookie you must provide all the options for that cookie."_

Ideally, we could change just the cookie names, and not have to provide all the other default options.

PR #182, opened 2023-01-23
Release 0.4.0

BracketJohn:
This PR is used to release `0.4.0` versions of `next-auth`.


BracketJohn:
Hey @Q16solver 👋 

Thanks for the remark - we do not need at least `4.8.18` as this fixes a bunch of regressions that `next-auth` themselves had. So I'll relax it a bit and set it to `^4.8.18`. Checkout #192

PR #199, opened 2023-01-31
fix call to next-auth AuthHandler

shaharmor:
The API changed to 2 parameters instead of 1: https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/core/index.ts#L266

Closes # .

Checklist:

### DIFF
--- a/src/runtime/server/services/nuxtAuthHandler.ts
+++ b/src/runtime/server/services/nuxtAuthHandler.ts
@@ -187,7 +187,7 @@ export const NuxtAuthHandler = (nuxtAuthOptions?: AuthOptions) => {
     // 1. Assemble and perform request to the NextAuth.js auth handler
     const nextRequest = await getInternalNextAuthRequestData(event)
 
-    const nextResult = await AuthHandler(nextRequest, options})
+    const nextResult = await AuthHandler(nextRequest, options)
 
     // 2. Set response status, headers, cookies
     if (nextResult.status) {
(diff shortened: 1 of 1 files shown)


src/runtime/server/services/nuxtAuthHandler.ts
```diff
@@ -187,10 +187,7 @@ export const NuxtAuthHandler = (nuxtAuthOptions?: AuthOptions) => {
     // 1. Assemble and perform request to the NextAuth.js auth handler
     const nextRequest = await getInternalNextAuthRequestData(event)
 
-    const nextResult = await AuthHandler({
-      req: nextRequest,
-      options
-    })
+    const nextResult = await AuthHandler(nextRequest, options})
```
BracketJohn, line 190: trailing closing bracket

BracketJohn (changes requested):
Hey @shaharmor 🏴󠁧󠁢󠁷󠁬󠁳󠁿 

Thanks for the contribution.

Unless I'm missing something, we cannot merge this though: You are referencing the current `main` branch-code, this however is wrong. We are on `^4.18.8`. If you look at the code associated with that, you can see that the interface is still the "old" one, see: https://github.com/nextauthjs/next-auth/blob/c385cf8c7cd9428a5a513e803ab1c05e0ca77354/packages/next-auth/src/core/index.ts#L80-L82

You can get to this place by browsing the releases and then viewing the files associated with each release: https://github.com/nextauthjs/next-auth/releases

Was there anything that didn't work / trip you up and prompted you to open this PR?

src/runtime/server/services/nuxtAuthHandler.ts
shaharmor: fixed

shaharmor:
@BracketJohn honestly, I'm not sure :)

I think I just had a very weird cache of next-auth which caused this issue.
What happened was that I didn't get the callbacks/events to trigger unless I did this change. But testing again it works so we can ignore this

PR #201, opened 2023-02-01
fix: use `route.matched` to determine if 404 was hit

### LINKED ISSUE #185: The option "allow404WithoutAuth" does not work if you have a route with a placeholder. Protected pages potentially leak.
### Environment

Nuxi 3.0.0                                                                                                   17:14:08
                                                                                                             17:14:09
RootDir: /app                                                                                                17:14:09
Nuxt project info:                                                                                           17:14:09

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.19.0`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `npm@9.3.1`
- Builder:          `vite`

### Reproduction

Add a new route in the pages folder with a local auth middleware.
[language]
.. protected
 ....index.vue
```
<template>
  <div>
    <h1>Protected</h1>
  </div>
</template>

<script lang="ts" setup>
definePageMeta({ middleware: 'auth' })
</script>
```

**Nuxt.config.ts**
```
  auth: {
    enableGlobalAppMiddleware: false,
  },`
```

In the middleware we have this code:
```
  if (authConfig.globalMiddlewareOptions.allow404WithoutAuth) {
    const matchedRoute = router.getRoutes().find((route) => route.path === to.path);
```

`to.path` is in this example `/de/protected`
`route.path` is `/:language/protected`

The is no match and the auth middleware does not redirect and show the protected page.

### Describe the bug

If you use a local middleware and protect a route with a placeholder, the page will be displayed.
Adding  **allow404WithoutAuth: true** but there is a high chance that someone exposes protected pages with this check.
**Nuxt.config.ts**
```
  auth: {
    enableGlobalAppMiddleware: false,
    globalMiddlewareOptions: {
      allow404WithoutAuth: true,
    },
  },`
```

### Additional context

_No response_

### Logs

_No response_

BracketJohn:
Closes #185

This PR: replaces the custom matching logic to the already existing one  of vue-router (big win).

This closes #185 as `vue-router` can correctly match dynamic urls like `/:language/protected` to `/de/protected`

### DIFF
--- a/playground/nuxt.config.ts
+++ b/playground/nuxt.config.ts
@@ -1,7 +1,7 @@
 export default defineNuxtConfig({
   modules: ['../src/module.ts'],
   auth: {
-    enableGlobalAppMiddleware: false,
+    enableGlobalAppMiddleware: true,
     defaultProvider: undefined
   }
 })
--- a/playground/nuxt.config.ts
+++ b/playground/nuxt.config.ts
@@ -1,7 +1,7 @@
 export default defineNuxtConfig({
   modules: ['../src/module.ts'],
   auth: {
-    enableGlobalAppMiddleware: true,
+    enableGlobalAppMiddleware: false,
     defaultProvider: undefined
   }
 })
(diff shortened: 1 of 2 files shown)


playground/nuxt.config.ts
```diff
@@ -1,7 +1,7 @@
 export default defineNuxtConfig({
   modules: ['../src/module.ts'],
   auth: {
-    enableGlobalAppMiddleware: true,
+    enableGlobalAppMiddleware: false,
```
BracketJohn, line 4:
```suggestion
    enableGlobalAppMiddleware: true,
```

PR #206, opened 2023-02-03
Documentation update: Update 4.custom-session-data.md

### LINKED ISSUE #148: Custom fields in session or user
Hi, opening a new issue since #61 was closed, but I am not sure if it's possible now to inject additional data or not yet, I have been trying to follow the documentation and also the example provided by @JoaoPedroAS51 but nothing seems to work.

In my case I am using the credentials provider, and the big problem for me is where to put the data from backend response to get it in the callbacks.

If I log token, account, profile in the callbacks they seem empty.

image
image

This is the ~/server/api/auth/[...].ts
```
import CredentialsProvider from 'next-auth/providers/credentials'
import { NuxtAuthHandler } from '#auth'

export default NuxtAuthHandler({
  providers: [
    // @ts-expect-error Import is exported on .default during SSR, so we need to call it this way. May be fixed via Vite at some point
    CredentialsProvider.default({
      // The name to display on the sign in form (e.g. 'Sign in with...')
      name: 'Credentials',
      // The credentials is used to generate a suitable form on the sign in page.
      // You can specify whatever fields you are expecting to be submitted.
      // e.g. domain, username, password, 2FA token, etc.
      // You can pass any HTML attribute to the <input> tag through the object.
      credentials: {
        username: { label: 'Username', type: 'text', placeholder: 'jsmith' },
        password: { label: 'Password', type: 'password' },
      },
      profile(profile, tokens) {
        console.log('profile', profile)
        console.log('tokens', tokens)
        return {
          ...profile,
        }
      },

      async authorize(credentials) {
        // You need to provide your own logic here that takes the credentials
        // submitted and returns either a object representing a user or value
        // that is false/null if the credentials are invalid.
        // e.g. return { id: 1, name: 'J Smith', email: 'jsmith@example.com' }
        // You can also use the `req` object to obtain additional parameters
        // (i.e., the request IP address)

        // console.log(credentials)
        // console.log(req.body)
        // const user = { id: '1', name: 'J Smith', email: 'jsmith@example.com' }
        const token = Buffer.from(`${credentials.username}:${credentials.password}`).toString('base64')
        // console.log(token)
        const res = await fetch('http://localhost/backend/login', {
          method: 'post',
          headers: {
            'Content-Type': 'application/json',
            'Authorization': `Basic ${token}`,
          },
        })
        const data = await res.json()
        // If no error and we have user data, return it
        if (res.ok && res) {
          return {
            name: data.data.user.username,
            email: data.data.user.email,
            image: data.data.token,
            test: 'blah',
            aaa: 'aaa',
          }
        }
        else {
          // If you return null then an error will be displayed advising the user to check their details.
          return null

          // You can also Reject this callback with an Error thus the user will be sent to the error page with the error message as a query parameter
        }
      },
    }),
  ],
  callbacks: {
    // Callback when the JWT is created / updated, see https://next-auth.js.org/configuration/callbacks#jwt-callback
    async jwt({ token, account, profile }) {
      // Persist the OAuth access_token and or the user id to the token right after signin
      console.log('account', account)
      console.log('profile', profile)
      console.log('token', token)
      // if (account) {
      //   token.accessToken = account.token
      //   token.id = profile.id
      // }
      return token
    },
    // Callback whenever session is checked, see https://next-auth.js.org/configuration/callbacks#session-callback
    async session({ session, token, user }) {
      // Send properties to the client, like an access_token and user id from a provider.
      console.log('token', token)
      // session.accessToken = token.accessToken
      // session.user.id = token.id
      return session
    },
  },
})

```

_Originally posted by @jorgv in https://github.com/sidebase/nuxt-auth/issues/61#issuecomment-1373557623_

jorgv:
Updating the custom session data recipe documentation based on #148

Closes #148 .

Checklist:

### DIFF
--- a/docs/content/5.recipes/4.custom-session-data.md
+++ b/docs/content/5.recipes/4.custom-session-data.md
@@ -1,6 +1,6 @@
 # Custom Session Data
 
-This guide explains how to add custom data to the user session, in this example we will continue to use the [Strapi Recipe](/nuxt-auth/recipes/strapi)  but it can easily be applied to any OAuth provider.
+This guide explains how to add custom data to the user session.
 
 To expand / change / customize the session data you need to to follow the following steps:
 1. Return the custom data in the Authorize function
@@ -21,27 +21,23 @@ export default NuxtAuthHandler({
     CredentialsProvider.default({
     ...
       async authorize(credentials: any) {
-        const response = await $fetch<any>(
-          `${process.env.STRAPI_BASE_URL}/auth/local/`,
-          {
-            method: "POST",
-            body: JSON.stringify({
-              identifier: credentials.username,
-              password: credentials.password,
-            }),
-          }
-        ).catch((e) => {
-          throw createError({ statusCode: 400, statusMessage: "Bad Request" });
-          console.log(e);
-        });
+        const data = {
+          user: {
+            name: 'John Doe',
+            email: 'john@email.com',
+            id: 1,
+            role: 'admin'
+          },
+          token: '123456890',
+        }
 
-        if (response.user) {
+        if (data.user) {
           const u = {
-            id: response.user.id,
-            name: response.user.username,
-            email: response.user.email,
-            access_token: response.jwt, // additional field
-            role: response.user.role
+            id: data.user.id,
+            name: data.user.username,
+            email: data.user.email,
+            access_token: data.token, // additional field
+            role: response.user.role // additional field
           };
           return u;
         } else {
@@ -113,7 +109,7 @@ const token = await getToken({ event })
 console.log(token.jwt);
 ```
 
-Please Notice that with this approach there is no need to use profile function or override the types next-auth.
+With this approach `data` will automatically be correctly typed.
 
 
 Thanks to 
--- a/docs/content/5.recipes/4.custom-session-data.md
+++ b/docs/content/5.recipes/4.custom-session-data.md
@@ -60,7 +60,7 @@ First, you must add jwt and session callbacks to [the `NuxtAuthHandler`](/nuxt-a
 ```ts
 // file: ~/server/api/auth/[...].ts
 export default NuxtAuthHandler({
-...
+  // ... remainder of your config ...
   callbacks: {
     // Callback when the JWT is created / updated, see https://next-auth.js.org/configuration/callbacks#jwt-callback
     jwt: async ({token, user}) => {
--- a/docs/content/5.recipes/4.custom-session-data.md
+++ b/docs/content/5.recipes/4.custom-session-data.md
@@ -1,57 +1,121 @@
 # Custom Session Data
 
-This guide explains how to add custom data to the user session.
+This guide explains how to add custom data to the user session, in this example we will continue to use the [Strapi Recipe](/nuxt-auth/recipes/strapi)  but it can easily be applied to any OAuth provider.
 
 To expand / change / customize the session data you need to to follow the following steps:
-1. Add callbacks to the `NuxtAuthProvider` to alter the session data
-2. Update the TypeScript interfaces of `nuxt-auth` to get accurate typing in your application
+1. Return the custom data in the Authorize function
+2. Add callbacks to the `NuxtAuthProvider` to alter the session data
 
-In the following example we'll add a new field to the session: `accessToken`. This `accessToken` will the `access_token` we receive from an imaginary OAuth provider.
+In the following example we'll add 2 new fields to the session: `accessToken` and `role`.
 
-## Adding Callback to Alter Session Data
+## Return the custom data in the Authorize function
+
+
+```ts
+// file: ~/server/api/auth/[...].ts
+
+export default NuxtAuthHandler({
+  secret: config.NUXT_SECRET,
+  providers: [
+    // @ts-expect-error You need to use .default here for it to work during SSR. May be fixed via Vite at some point
+    CredentialsProvider.default({
+    ...
+      async authorize(credentials: any) {
+        const response = await $fetch<any>(
+          `${process.env.STRAPI_BASE_URL}/auth/local/`,
+          {
+            method: "POST",
+            body: JSON.stringify({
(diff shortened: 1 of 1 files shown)


docs/content/5.recipes/4.custom-session-data.md
```diff
@@ -1,57 +1,121 @@
...
+console.log(token.jwt);
+```
+
+Please Notice that with this approach there is no need to use profile function or override the types next-auth.
+
 
-Thanks to [@JoaoPedroAS51 from Github](https://github.com/JoaoPedroAS51) for providing this information here: https://github.com/sidebase/nuxt-auth/issues/61#issuecomment-1330747199
+Thanks to 
+- [@JoaoPedroAS51 from Github](https://github.com/JoaoPedroAS51) for his contribution here: https://github.com/sidebase/nuxt-auth/issues/61#issuecomment-1330747199
+- [@pvlastaridis from Github](https://github.com/pvlastaridis) for his contribution here: https://github.com/sidebase/nuxt-auth/issues/148#issuecomment-1407490921
```
BracketJohn, line 121: you can also thank yourself here (:

docs/content/5.recipes/4.custom-session-data.md
```diff
@@ -1,57 +1,121 @@
...
+
+```
+import { getToken } from '#auth'
+
+const token = await getToken({ event })
+
+console.log(token.jwt);
+```
+
+Please Notice that with this approach there is no need to use profile function or override the types next-auth.
```
BracketJohn, line 116: oh wow, that's nice, let's phrase it differently though as futurre readers will not know about the past approach:
```md
With this approach `data` will automatically be correctly typed.
```

BracketJohn (changes requested):
Hey @jorgv 👋 

Thank you so much for taking me up on my question and helping out with our docs. Let's get this merged and help a lot of other people that have a similar problem.

I propose that we:
1. do not build on the Strapi recipe but instead keep this doc change contained to itself. I don't think that Strapi is necessary here to showcase what we want to do,
2. keep the example minimal by adding only 1 - 2 fields 

Instead, you can replace the strapi `$fetch` call with something like:
```ts
// You can replace this with any data, `$fetch` result, ... in your own code
const data = {
  user: {
    id: 1,
    role: 'admin'
  }
}
```

BracketJohn:
Let me know if I can support you further - would love to land this clarification (:

docs/content/5.recipes/4.custom-session-data.md
jorgv: Thank you, I will make the changes tomorrow

docs/content/5.recipes/4.custom-session-data.md
jorgv: No need, I am just adapting the solution to the docs. I am just grateful of the splendid work everyone is doing.

docs/content/5.recipes/4.custom-session-data.md
```diff
@@ -3,55 +3,115 @@
 This guide explains how to add custom data to the user session.
 
 To expand / change / customize the session data you need to to follow the following steps:
-1. Add callbacks to the `NuxtAuthProvider` to alter the session data
-2. Update the TypeScript interfaces of `nuxt-auth` to get accurate typing in your application
+1. Return the custom data in the Authorize function
```
BracketJohn, line 6:
```suggestion
1. Return the custom data in the `authorize` function
```

docs/content/5.recipes/4.custom-session-data.md
```diff
@@ -3,55 +3,115 @@
...
+```
+We can see the new fields in the return object `u` `role` and `access_token`. Now we need to pass them in the callbacks.
+
+## Add callbacks to the `NuxtAuthProvider` to alter the session data
 
 First, you must add jwt and session callbacks to [the `NuxtAuthHandler`](/nuxt-auth/configuration/nuxt-auth-handler):
 ```ts
 // file: ~/server/api/auth/[...].ts
 export default NuxtAuthHandler({
+...
```
BracketJohn, line 63:
```suggestion
  // ... remainder of your config ...
```

docs/content/5.recipes/4.custom-session-data.md
```diff
@@ -3,55 +3,115 @@
...
+      (session as any).role = token.role;
+      (session as any).uid = token.id;
+      return Promise.resolve(session);
+    },
   },
   ..., // other config
   providers: [...], // Your provider config
 })
 ```
+Notice that in the `jwt` callback we pass both the access_token and the role but in the session callback we leave out the access_token and we only pass the role since this will expose jwt on the frontend through the session object. This would be the recommendable approach, but you can pass it to the session if needed.
```
BracketJohn, line 86:
```suggestion
Notice that in the `jwt` callback we pass both the `access_token` and the `role` but in the session callback we leave out the `access_token` and we only pass the role since this would expose the token on the frontend through the session object. Exposing the `access_token` is not recommended, but you can pass it to the session if needed.
```

docs/content/5.recipes/4.custom-session-data.md
```diff
@@ -3,55 +3,115 @@
...
 
-## Updating TypeScript Interfaces for Accurate Typing
+After that you can access the new fields in the data portion of the session object:
 
-```ts
-// file: ~/next-auth.d.ts
-import { DefaultSession } from "next-auth"
+```
```
BracketJohn, line 90:
```suggestion
```ts
```

docs/content/5.recipes/4.custom-session-data.md
```diff
@@ -3,55 +3,115 @@
...
+
+The access_token can be access as:
+
+```
+import { getToken } from '#auth'
+
+const token = await getToken({ event })
+
+console.log(token.jwt);
+```
```
BracketJohn, line 110:
```suggestion
```

BracketJohn (approved):

PR #215, opened 2023-02-06
#200 Refresh Token Rotation does not update cookie if called server s…

### LINKED ISSUE #200: Refresh Token Rotation does not update cookie if called server side.
### Environment

_No response_

### Reproduction

_No response_

### Describe the bug

I have spent hours on making the **Refresh Token Rotation** work based on the existing example. 
https://sidebase.io/nuxt-auth/recipes/directus
and
https://next-auth.js.org/tutorials/refresh-token-rotation

As a test scenario, I have integrated an OAuth server and was able to authenticate against it.
I set the AccessToken lifetime to 10 seconds. After 10 seconds I did a hard refresh in the browser to simulate a server side request on Nuxt. This correctly fetches a new **AccessToken** and a **RefreshToken**.

Unfortunatly this does not update the JWT cookie in the browser, the old cookie is sent along.

After 10 seconds I did another hard refresh. This results in an error, because another refresh happens but based on an invalidated refresh token from the previous cookie.

After debugging I figured out, that the token is refreshed correctly. This happens in the route middleware where `useSession()` is called. There a fetch to the session server endpoint happens and the token is updated correctly.

To make it work Iextracted the the `setCookie` header from the request and passed it  along with the server side request.

In order to support refresh token rotation I propose the following fix:

File:
https://github.com/sidebase/nuxt-auth/blob/6696abdaed350b066ab74c9541e88c774743255a/src/runtime/composables/useSession.ts#L203

```
  const headers = await getRequestCookies(nuxt)

  return _fetch<SessionData>(nuxt, 'session', {
    onResponse: ({ response }) => {
      const sessionData = response._data

      // Pass the new cookie to the server side request   
      if (process.server) {
        if (response.headers.get('set-cookie')) {
          const setCookieValue = response.headers.get('set-cookie')
          appendHeader(nuxt.ssrContext.event, 'set-cookie', setCookieValue)
        }
      }

      data.value = isNonEmptyObject(sessionData) ? sessionData : null
      loading.value = false

      if (required && status.value === 'unauthenticated') {
        return onUnauthenticated()
      }

      return sessionData
    },
```
Basically I'm just appending the updated `setCookie` header to the existing SSR request. This will updated the cookie correctly.

I can also make a pull request as soon as someone validated my proposal.

### Additional context

_No response_

### Logs

_No response_

vanling:
Copied the small proposal for a fix for #200 from @ayalon, applied his fix locally and the JWT cookie was updated correctly everytime.

I have no understanding of how 90% of this all works, so this fix might require some more testing with other providers.

Closes #200  .

Checklist:
- [x] manually checked my feature 

### DIFF
--- a/src/runtime/composables/useSession.ts
+++ b/src/runtime/composables/useSession.ts
@@ -4,6 +4,7 @@ import { callWithNuxt } from '#app'
 import { readonly } from 'vue'
 import type { ComputedRef, Ref } from 'vue'
 import type { NuxtApp } from '#app'
+import { appendHeader } from 'h3'
 import { getRequestURL, joinPathToApiURL, navigateToAuthPages } from '../utils/url'
 import { _fetch } from '../utils/fetch'
 import { isNonEmptyObject } from '../utils/checkSessionResult'
@@ -14,7 +15,6 @@ import type {
   SessionStatus
 } from './useSessionState'
 import { createError, useNuxtApp, useRuntimeConfig, useRequestHeaders } from '#imports'
-import { appendHeader } from 'h3'
 
 /**
  * Utility type that allows autocompletion for a mix of literal, primitiva and non-primitive values.
--- a/src/runtime/composables/useSession.ts
+++ b/src/runtime/composables/useSession.ts
@@ -208,11 +208,9 @@ const getSession = async (getSessionOptions?: GetSessionOptions) => {
       // Add any new cookie to the server-side event for it to be present on the app-side after
       // initial load, see sidebase/nuxt-auth/issues/200 for more information.
       if (process.server) {
-        if (response.headers.get('set-cookie') && nuxt.ssrContext) {
-          const setCookieValue = response.headers.get('set-cookie') || ''
-          if (setCookieValue) {
-            appendHeader(nuxt.ssrContext.event, 'set-cookie', setCookieValue)
-          }
+        const setCookieValue = response.headers.get('set-cookie')
+        if (setCookieValue && nuxt.ssrContext) {
+          appendHeader(nuxt.ssrContext.event, 'set-cookie', setCookieValue)
         }
       }
 
--- a/src/runtime/composables/useSession.ts
+++ b/src/runtime/composables/useSession.ts
@@ -14,6 +14,7 @@ import type {
   SessionStatus
 } from './useSessionState'
 import { createError, useNuxtApp, useRuntimeConfig, useRequestHeaders } from '#imports'
+import { appendHeader } from 'h3'
 
 /**
  * Utility type that allows autocompletion for a mix of literal, primitiva and non-primitive values.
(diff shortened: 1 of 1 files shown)


BracketJohn (changes requested):
Hey @vanling 👋 

Thanks for the PR and taking up this work 💯 


> I have no understanding of how 90% of this all works, so this fix might require some more testing with other providers.

I think for now this is fine: It only adds some cookie setting, so the change should work independent of different providers.

One nit and we're good to go

BracketJohn:
Also, please resolve the open TS errors:
image

Test locally by running:
```sh
npm run types
```

vanling:
That is going to be an adventure 😂 I will have a try at it tonight =)

BracketJohn:
Sure thing. Should be straightforward and if it isn't let me know and I can give you a hand. Basically there's two errors:
1. `nuxt.ssrContext` may be undefined -> resolve it by checking if it exists
2. `setCookieValue` may be undefined -> resolve this by also checking if it exists before assigning it

So, easy enough 😄

ayalon:
I have fixed the typescript error. Thanks for the collaboration on this issue. Highly appreciated.

vanling:
Thanks @ayalon I am happy you joined in this one :D did not want to steal al your credits ;)

ayalon:
The tests legitimately failed. Awesome coverage. I have fixed the missing import. Let's see if this works.

src/runtime/composables/useSession.ts
```diff
@@ -204,6 +205,17 @@ const getSession = async (getSessionOptions?: GetSessionOptions) => {
...
+      // Add any new cookie to the server-side event for it to be present on the app-side after
+      // initial load, see sidebase/nuxt-auth/issues/200 for more information.
+      if (process.server) {
+        if (response.headers.get('set-cookie') && nuxt.ssrContext) {
+          const setCookieValue = response.headers.get('set-cookie') || ''
+          if (setCookieValue) {
+            appendHeader(nuxt.ssrContext.event, 'set-cookie', setCookieValue)
+          }
+        }
+      }
```
BracketJohn, line 217: I think we can improve this a bit:
```suggestion
      // Add any new cookie to the server-side event for it to be present on the app-side after
      // initial load, see sidebase/nuxt-auth/issues/200 for more information.
      if (process.server) {
        const setCookieValue = response.headers.get('set-cookie')
        if (setCookieValue && nuxt.ssrContext) {
          appendHeader(nuxt.ssrContext.event, 'set-cookie', setCookieValue)
        }
      }
```

BracketJohn (changes requested):

BracketJohn:
> The tests legitimately failed. Awesome coverage. I have fixed the missing import. Let's see if this works.

thanks, this is an awesome compliment. Just reviewed tyour code (:

vanling:
Changed the import order, ran lint + types local. Should be ok now.

BracketJohn (approved):

PR #219, opened 2023-02-09
feat: Add getServerSessionExternal

### LINKED ISSUE #135: RFC: `getServerSession` support for Websockets and other non-H3Event requests
### Describe the feature

It'd be nice if `getServerSession` could be called with non-`H3Event` payloads, to allow all standardized `req` / `res` sources to get the server session. A prominent example of this are `websocket`s.

### Additional information

This arose in #131. 

@q16solver's original remark:

> createContext actually receives the H3Event from the first argument unless you're using websockets as well, but I do agree that getServerSession's api from nuxt auth should be fixed up and have the same structure as next auth, I shouldn't need to pass in the whole event argument so it can grab the session, it should only need req and res like next auth, because for websockets, there is no H3Event, but I should still be able to have authenticated websockets and be able to grab the session.

> This is especially painful when I'm trying to implement auth as middlewares in trpc since I need to propagate the entire event object throughout the context, hopefully this can be fixed c:

> @ngajhede I think we can probably keep this issue open and perhaps put a different tag like enhancement to improve the api, what do you think @BracketJohn?

Q16solver:
Closes #135 .

Had a stab at just refactoring the existing getServerSession to only accept requests, not sure if this is any good, will also do a bit of testing as well to see if it works

Checklist:


Q16solver:
Honestly could probably merge them together and just have 2 auth handlers and replace the functions with the req version of it provided that it behaves the same way, probably need to have a deeper look once I have a bit of time

Q16solver:
Closing this to open a PR from a diff branch so I don't ruin myself when syncing main branch

PR #224, opened 2023-02-09
feat: Add getServerSessionExternal

### LINKED ISSUE #135: RFC: `getServerSession` support for Websockets and other non-H3Event requests

Q16solver:
Closes #135  .

Had a stab at just refactoring the existing getServerSession to only accept requests, not sure if this is any good, will also do a bit of testing as well to see if it works

Checklist:


Q16solver:
@BracketJohn This was actually not nearly as bad as I thought, it's actually much simpler, it makes sense that for getServerSessionExternal (I kind of just called it this since I don't have a better name and thought of this on the fly), that it really should just only be for grabbing the session, and not handle any other logic, all the logging in etc, should already be handled by the original event handler

BracketJohn:
Thanks @Q16solver for this - sorry I don't have much time at the moment so it'll take some time for me to get around to review this - hope you can understand (:

PR #237, opened 2023-02-13
Make sure SessionRefreshPeriodically is run on client only

vanling:
Closes #236 .

Checklist:
- [x] issue number linked above after pound (`#`)
- [x] manually checked my feature + lint + types
- [x] testing not applicable
- [x] screenshot not applicable

PR #250, opened 2023-02-20
Release 0.4.2

zoey-kaiser:
Releases version 0.4.2 with PR change #237

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sidebase/nuxt-auth",
-  "version": "0.4.1",
+  "version": "0.4.2",
   "license": "MIT",
   "type": "module",
   "exports": {


BracketJohn:
Please do:
- an alpha release
- install and tesst it in the nuxt auth example
- if everything is good: do real release

zoey-kaiser:
> an alpha release

Released 0.4.2-alpha.1

> install and tesst it in the nuxt auth example

Added it to `nuxt-auth-example`
- Started Application
- Logged in through credentials
- Checked all routes
- Logged out

> if everything is good: do real release

I think we are readz to proceed with the release!

BracketJohn:
let's go!

BracketJohn (approved):

PR #253, opened 2023-02-21
fix(cookies): Handle next-auth cookies on the server

### LINKED ISSUE #238: signIn does not set cookies when called on the server
### Environment

nuxt-auth: 0.4.1
provider: OAuth

```
- Operating System: `Windows_NT`
- Node Version:     `v16.18.1`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `2.2.1`
- Package Manager:  `pnpm@7.25.1`
- Builder:          `vite`
- User Config:      `ssr`, `modules`, `build`, `components`, `typescript`, `vueuse`, `inkline`, `auth`, `runtimeConfig`
- Runtime Modules:  `@vueuse/nuxt@9.11.1`, `@inkline/nuxt@3.1.4`, `@sidebase/nuxt-auth@0.4.1`
- Build Modules:    `-`
```

### Reproduction

* Configure the Google OAuth provider
* Set it as the `defaultProvider`
* Visit a protected page (you are redirected to Google directly)
* Login with Google

You should have been redirected to the signin page with an OAuth error and there should be a `checks.state argument is missing` message in the server console.

### Describe the bug

The `signIn` function sends a POST request to the `/api/auth/signin/PROVIDER` url.

With the default OAuth configuration, this request sets cookies which are required to validate the OAuth result (for the state/pkce checks).

These cookies are not propagated to the client when the call is made on the server, resulting in a `checks.state argument is missing` error from `next-auth`.

### Additional context

https://github.com/sidebase/nuxt-auth/blob/c231ae839dc180c03a67bdf8ca44df1aa58da8b1/src/runtime/composables/useSession.ts#L150-L161

### Logs

_No response_

### LINKED ISSUE #244: Redirect to default provider's login page immediately
### Ask your question

How do I make a protected page redirect to the provider's login page directly and not to the provider selection page?  For example, I'm not yet authenticated and I will just enter a protected page in the browser like http://localhost:5000/profile.  Since I only have 1 provider, I want to be redirected immediately to the login page without having the following page show up first.

<kbd>provider_select</kbd>

Currently, my `..\server\api\auth\[...].ts` is configured as:

```
import KeycloakProvider from 'next-auth/providers/keycloak'
import { NuxtAuthHandler } from "#auth"

export default NuxtAuthHandler({
    // secret needed to run nuxt-auth in production mode (used to encrypt data)
    secret: process.env.NUXT_SECRET,
    providers: [
        KeycloakProvider.default({
            clientId: process.env.AUTH_CLIENT_ID,
            clientSecret: process.env.AUTH_CLIENT_SECRET,
            issuer: process.env.AUTH_ISSUER
        })
    ]
})
```

Also, I have enabled the global authentication middleware in `nuxt.config.ts`

```
// https://v3.nuxtjs.org/api/configuration/nuxt.config
  modules: ['@sidebase/nuxt-auth'],
  auth: {
    origin: process.env.ORIGIN,
    enableGlobalAppMiddleware: true
  }
})
```

### Additional information

_No response_

Desuuuu:
The goal of this PR is to handle `next-auth` cookies properly during SSR.

When rendering pages on the server, multiple calls to `next-auth` can be made (session, csrf, signin, etc) and these calls usually expect/set cookies (`next-auth.callback-url`, `next-auth.csrf-token`, `next-auth.state`, `next-auth.pkce.code_verifier`, ...).

Currently, these cookies are not passed between calls and are not passed to the browser, resulting in these issues:
* When no CSRF cookie is present on the browser (first page load): a redirection to the signin page, even though `defaultProvider` is set, due to a CSRF mismatch.
* When a signIn with a provider (or the default provider) is issued on the server, a `checks.state argument is missing` error, due to missing/outdated checks cookies.

With this change, `next-auth` cookies are passed from request to request on the server and are ultimately passed to the browser.

The cookie parsing code is adapted from [tough-cookie](https://github.com/salesforce/tough-cookie) and the `Set-Cookie` splitting code is from [set-cookie-parser](https://github.com/nfriedly/set-cookie-parser).

Closes #238, closes #244.

Checklist:

### DIFF
--- a/src/runtime/composables/useSession.ts
+++ b/src/runtime/composables/useSession.ts
@@ -4,7 +4,6 @@ import { callWithNuxt } from '#app'
 import { readonly } from 'vue'
 import type { ComputedRef, Ref } from 'vue'
 import type { NuxtApp } from '#app'
-import { appendHeader } from 'h3'
 import { getRequestURL, joinPathToApiURL, navigateToAuthPages } from '../utils/url'
 import { _fetch } from '../utils/fetch'
 import { isNonEmptyObject } from '../utils/checkSessionResult'
@@ -204,15 +203,6 @@ const getSession = async (getSessionOptions?: GetSessionOptions) => {
   return _fetch<SessionData>(nuxt, 'session', {
     onResponse: ({ response }) => {
       const sessionData = response._data
-
-      // Add any new cookie to the server-side event for it to be present on the app-side after
-      // initial load, see sidebase/nuxt-auth/issues/200 for more information.
-      if (process.server) {
-        const setCookieValue = response.headers.get('set-cookie')
-        if (setCookieValue && nuxt.ssrContext) {
-          appendHeader(nuxt.ssrContext.event, 'set-cookie', setCookieValue)
-        }
-      }
 
       data.value = isNonEmptyObject(sessionData) ? sessionData : null
       loading.value = false
--- a/src/runtime/utils/cookies.ts
+++ b/src/runtime/utils/cookies.ts
@@ -1,0 +1,426 @@
+import { getResponseHeader, setResponseHeader } from 'h3'
+import type { H3Event } from 'h3'
+
+interface RawCookie {
+  name: string
+  value: string
+  expiration?: Date
+  raw: string
+}
+
+// eslint-disable-next-line no-control-regex
+const CONTROL_CHARS = /[\x00-\x1F]/
+const TERMINATORS = ['\n', '\r', '\0']
+// eslint-disable-next-line no-control-regex
+const DATE_DELIM = /[\x09\x20-\x2F\x3B-\x40\x5B-\x60\x7B-\x7E]/
+const MONTH_TO_NUM: Record<string, number> = {
+  jan: 0,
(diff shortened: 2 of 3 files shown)


zoey-kaiser:
Hi @Desuuuu,

Thanks for taking the initiative and opening this PR 😍
I will do a first review and then pass it on to @BracketJohn. I think these changes are very much needed so ill work on prioritizing it!

BracketJohn (commented):
Hey @Desuuuu 👋 

Thanks for this PR!

I think:
- we should look into recent Nuxt 3.2 changes that changed cookie behavior and see if they already alleviate this issue
- we should not manually (re-)implement cookie parsing here - why:
    - can we not use `useRequestHeaders(['cookies'])`?
    - if we must further parse the cookies: why not pull in a library instead of copy pasting it's code into here?

Again, thanks for the contribution!

BracketJohn:
Thanks! Then I propose to:
- have a separate PR for the Nuxt 3.3 update for a partial fix of headers
- add everything as dependencies / use from `h3` that could be used from there

I think that should clean up this PR a lot (:

PR #304, opened 2023-04-04
Add guest mode

BracketJohn:
This PR:
- adds guest mode
- adds guest mode docs


develth:
White this Line allows that guest see the page, when it is configured that it should not happen (`unauthenticatedOnly = false`) 
[ see case](https://github.com/sidebase/nuxt-auth/blob/2374f40d55082e27f13120a33b028c4d5fa252d6/src/runtime/middleware/auth.ts#L58)

PR #332, opened 2023-04-17
docs: add azure AD example

thomasfjordside:
adds a  azure AD NuxtAuthHandler example to recipes documentation section.

### DIFF
--- a/docs/content/5.recipes/8.azure-ad-example.md
+++ b/docs/content/5.recipes/8.azure-ad-example.md
@@ -1,0 +1,84 @@
+# Azure AD
+
+This section gives an example of a configuration of `NuxtAuthHandler` for use with Azure AD and shows an implemention of token refresh for Azure AD JWT.
+
+The code below is an example and you will need to adapt it to your own setup:
+```js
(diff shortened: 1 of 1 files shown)


BracketJohn (approved):
looks good, thanks for the contribution!

JeremyQtweb:
I'm trying to build out something with this and in 0.5.0 and 0.6.0 I get the following TS errors using this AzureADProvider example. I don't expect the auth to work with config yet, but I didn't expect the problems outlined here.
This is an empty project, it literally just has the auth in it, the plan for it being a layer for other projects to pull in when needed (if that will even work :P).

image

Any guidance would be appreciated. Keen to expand on the AzureAD recipe when I figure out how to get it working!

thomasfjordside:
@JeremyQtweb Sorry for the late reply, it must hidden itself, somewhere. Sadly I'm not that sharp in typescript. My example is from a nuxt project running with js files, where this does not cause issues. I have not tried the recipe in 0.6.0. Hopefully you have already gotten help elsewhere. Otherwise the sidebase discord is helpful.

PR #364, opened 2023-05-03
Adding basic `DevTools` integration

zoey-kaiser:
WIP PR. Added a simple DevTools UI that show very little data as a POC. Still need to figure out how to:

- Access `useAuth` composable to display session data that is useful for debugging and not just the current config
- Decide what else should be shown on the DevTools page
- How the selected providers when using `authJS` provider


zoey-kaiser:
Closing for now. Can we re-opened or created in the future!

PR #375, opened 2023-05-10
fix: #356 - Extend/change local signIn credentials type

### LINKED ISSUE #356: v0.6.0 - Extend/change local signIn credentials type?
### Ask your question

I'm using v0.6.0-alpha.2 with the local method of signIn. It seems to be working as expected. However the backend needs the submission to use email as the identity field and requires additional fields for verification, e.g. form and recaptcha tokens as below.
image
While this works functionally I'm getting typescript errors as the submission does not conform to the expected type. I can't currently see any way to extend or modify the signIn "Credentials" type unless I'm missing something.
What would be the suggested way to handle this since there doesn't actually seem to be any functional dependency on the signIn call that would require only username and password as credentials?

### Additional information

_No response_

Danielwinkelmann:
Closes #356 .

Checklist:
- [x] manually checked my feature
- [x] testing not applicable
- [x] screenshot not applicable


BracketJohn (approved):
Todo for later: improve docs -> I just checked out https://sidebase.io/nuxt-auth/v0.6/application-side/session-access-and-management and saw that the local part is a bad copy-pasta

Thanks @Danielwinkelmann ❤️ 🤝

PR #377, opened 2023-05-10
Fix mispellings of "immediately".

KyleSmith0905:
A small commit to fix misspellings of immeadiatly. This should not have any change around the app as I have only changed text in docs and JSDoc. So it could be fixed immeadiatly and doesn't need tests :D

Checklist:
- [x] checking not applicable
- [x] testing not applicable
- [x] screenshot not applicable


BracketJohn (approved):
Thanks - great contribution ❤️ ❤️

PR #409, opened 2023-05-24
fix: useAuthState setting useless cookie

### LINKED ISSUE #403: `useAuthState` adds useless `set-cookie` headers
### Environment

------------------------------
- Operating System: Darwin
- Node Version:     v18.13.0
- Nuxt Version:     3.5.0
- Nitro Version:    2.4.1
- Package Manager:  pnpm@8.4.0
- Builder:          vite
- User Config:      build, runtimeConfig, routeRules, css, typescript, components, schemaOrg, app, imports, modules, auth, hooks
- Runtime Modules:  @nuxtjs/tailwindcss@6.7.0, @formkit/nuxt@1.0.0-beta.11-0dd967c, nuxt-schema-org@2.1.3, @pinia/nuxt@0.4.11, @sidebase/nuxt-auth@0.6.0-beta.2
- Build Modules:    -
------------------------------

### Reproduction

_No response_

### Describe the bug

`useAuthState` uses Nuxt's `useCookie`-Hook with the default option set to `() => null`:

https://github.com/sidebase/nuxt-auth/blob/d8cb19037307c81c10b29ebaec25be8cf329db23/src/runtime/composables/local/useAuthState.ts#L20

which adds this header `set-cookie: auth:token=null; Max-Age=-1; Path=/; SameSite=Lax` to the main document response (preventing services like Vercel/Cloudflare from caching the page).

This should be an easy fix: [My fork](https://github.com/sidebase/nuxt-auth/commit/6b4f8fa98c9c07805fb687ce153d6b4ac602a27b)

### Additional context

_No response_

### Logs

_No response_

LouisHaftmann:
Closes #403

I did a bit more research and found out why Nuxt doesn't add the `Set-Cookie` header when the default value is undefined:

Nuxt's `useCookie` hook only adds a cookie header if the value of the cookie ref doesn't match the original value of the browser cookie ([source](https://github.com/nuxt/nuxt/blob/a672cd7a4291eb812d76b9841ec516cd8609c680/packages/nuxt/src/app/composables/cookie.ts#L46))

The cookie ref is initialized with the browser cookie's value or the default value passed to `useCookie` ([source](https://github.com/nuxt/nuxt/blob/a672cd7a4291eb812d76b9841ec516cd8609c680/packages/nuxt/src/app/composables/cookie.ts#L34))

The cookie is not set in the browser by default so `cookies[name]` returns `undefined`. The `useCookie` hook used in `useAuthState` has a default value of `null` and because Nuxt's `isEqual` function asserts that `null` does not equal `undefined`, the useless cookie header is added to the server response ([source](https://github.com/nuxt/nuxt/blob/a672cd7a4291eb812d76b9841ec516cd8609c680/packages/nuxt/src/app/composables/cookie.ts#LL46C11-L46C48))

This seems to be intended behavior of `useCookie` ([source](https://github.com/nuxt/nuxt/blob/a672cd7a4291eb812d76b9841ec516cd8609c680/packages/nuxt/src/app/composables/cookie.ts#L68))

Unless Nuxt changes it so `null` or `undefined` are just ignored, we should probably try to find another way of getting the token value without using `useCookie`.

My workaround just uses `undefined` as the default `auth:token` cookie value.

### DIFF
--- a/src/runtime/composables/local/useAuth.ts
+++ b/src/runtime/composables/local/useAuth.ts
@@ -51,7 +51,7 @@ const signOut: SignOutFunc = async (signOutOptions) => {
 
   const headers = new Headers({ [config.token.headerName]: token.value } as HeadersInit)
   data.value = null
-  rawToken.value = null
+  rawToken.value = undefined
 
   const { path, method } = config.endpoints.signOut
 
@@ -84,7 +84,7 @@ const getSession: GetSessionFunc<SessionData | null | void> = async (getSessionO
   } catch {
     // Clear all data: Request failed so we must not be authenticated
     data.value = null
-    rawToken.value = null
+    rawToken.value = undefined
   }
   loading.value = false
   lastRefreshedAt.value = new Date()
@@ -115,7 +115,7 @@ const signUp = async (credentials: Credentials, signInOptions?: SecondarySignInO
 
 interface UseAuthReturn extends CommonUseAuthReturn<typeof signIn, typeof signOut, typeof getSession, SessionData> {
   signUp: typeof signUp
-  token: Readonly<Ref<string | null>>
+  token: Readonly<Ref<string | undefined>>
 }
 export const useAuth = (): UseAuthReturn => {
   const {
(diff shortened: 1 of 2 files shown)


LouisHaftmann:
this will be fixed in the next Nuxt release

PR #422, opened 2023-06-02
fix: getServerSession returning null for authenticated users

f11y11:
Closes # .

Checklist:


f11y11:
I've implemeneted my own solution to fix getServerSession as told. I require additional confirmation since I'm unsure if you expect an identical response from next-auth.

PR #423, opened 2023-06-02
fix: accept boolean value for `globalAppMiddleware` option

LouisHaftmann:
This didn't work:

```ts
  auth: {
    globalAppMiddleware: true // this doesn't work
  },
})
```

because you needed to set `globalAppMiddleware` to

```ts
{
  isEnabled: true
}
```

But the type also allows a boolean value.


BracketJohn (approved):
Thanks! That will help many (:

PR #424, opened 2023-06-02
feat: protected route redirect overwrite per page

LouisHaftmann:
Adds a `navigateUnauthenticatedTo` option to `definePageMeta()` to allow for overwriting the default behavior of redirecting to the login page

### DIFF
--- a/src/runtime/middleware/auth.ts
+++ b/src/runtime/middleware/auth.ts
@@ -75,5 +75,9 @@ export default defineNuxtRouteMiddleware((to) => {
     const signInOptions: Parameters<typeof signIn>[1] = { error: 'SessionRequired', callbackUrl: determineCallbackUrl(authConfig, () => to.path) }
     // @ts-ignore This is valid for a backend-type of `authjs`, where sign-in accepts a provider as a first argument
     return signIn(undefined, signInOptions) as ReturnType<typeof navigateToAuthPages>
-  } else if (typeof metaAuth === 'object' && metaAuth.navigateUnauthenticatedTo) { return navigateTo(metaAuth.navigateUnauthenticatedTo) } else { return navigateTo(authConfig.provider.pages.login) }
+  } else if (typeof metaAuth === 'object' && metaAuth.navigateUnauthenticatedTo) {
+    return navigateTo(metaAuth.navigateUnauthenticatedTo)
+  } else {
+    return navigateTo(authConfig.provider.pages.login)
+  }
 })
--- a/src/runtime/middleware/auth.ts
+++ b/src/runtime/middleware/auth.ts
@@ -3,8 +3,8 @@ import { navigateToAuthPages, determineCallbackUrl } from '../utils/url'
 import { useAuth } from '#imports'
 
 type MiddlewareMeta = boolean | {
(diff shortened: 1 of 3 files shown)


BracketJohn (commented):
Sorry for taking so long to get around to this - was quite busy with my day job.

I like the feature - would just be good to also add some documentation here, same as in other PR! 

If you add aome we should be ready to go, thanks!

src/runtime/middleware/auth.ts
```diff
@@ -3,8 +3,23 @@ import { navigateToAuthPages, determineCallbackUrl } from '../utils/url'
 import { useAuth } from '#imports'
 
 type MiddlewareMeta = boolean | {
-  unauthenticatedOnly: true,
+  /** Whether to only allow unauthenticated users to access this page.
```
BracketJohn, line 6: CI is failing, should mostly be fixable by running `pnpm lint --fix` (:

BracketJohn (commented):
As in other PR of yours: Please move your docs-changes to the `v0.6` docs, thanks!

zoey-kaiser (dismissed):
LGTM!

src/runtime/middleware/auth.ts
```diff
@@ -60,7 +75,5 @@ export default defineNuxtRouteMiddleware((to) => {
     const signInOptions: Parameters<typeof signIn>[1] = { error: 'SessionRequired', callbackUrl: determineCallbackUrl(authConfig, () => to.path) }
     // @ts-ignore This is valid for a backend-type of `authjs`, where sign-in accepts a provider as a first argument
     return signIn(undefined, signInOptions) as ReturnType<typeof navigateToAuthPages>
-  } else {
-    return navigateTo(authConfig.provider.pages.login)
-  }
+  } else if (typeof metaAuth === 'object' && metaAuth.navigateUnauthenticatedTo) { return navigateTo(metaAuth.navigateUnauthenticatedTo) } else { return navigateTo(authConfig.provider.pages.login) }
```
valiafetisov, line 78: Let's increase readability 🙂 
```suggestion
  } else if (typeof metaAuth === 'object' && metaAuth.navigateUnauthenticatedTo) {
    return navigateTo(metaAuth.navigateUnauthenticatedTo)
  } else {
    return navigateTo(authConfig.provider.pages.login)
  }
```

src/runtime/middleware/auth.ts
```diff
@@ -3,8 +3,23 @@ import { navigateToAuthPages, determineCallbackUrl } from '../utils/url'
...
 
 type MiddlewareMeta = boolean | {
-  unauthenticatedOnly: true,
+  /** Whether to only allow unauthenticated users to access this page.
+   *
+   * Authenticated users will be redirected to `/` or the route defined in `navigateAuthenticatedTo`
+   *
+   * @default undefined
+   */
+  unauthenticatedOnly?: boolean,
```
valiafetisov, line 12: You did made this value to be optional, but you didn't set the default value anywhere. I suggest to keep `true` to be the default to not change previous expected behaviour

zoey-kaiser (approved):

PR #425, opened 2023-06-02
feat: getSession force refetch option

LouisHaftmann:
Adds a `force` option to `getSession`. Useful when token cookie has been updated by an external api route (not handled by nuxt-auth's `signIn`/`signOut`) because the token in `useAuthState` would still be `undefined`/`null` after manually calling a register api route for example.

### DIFF
--- a/docs/content/3.application-side/2.session-access-and-management.md
+++ b/docs/content/3.application-side/2.session-access-and-management.md
@@ -4,8 +4,6 @@
 Prior to v0.5.0 `useAuth()` was called `useSession()`.
 ::
 
-## `useAuth` Composable
-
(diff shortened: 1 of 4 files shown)


BracketJohn (dismissed):
Thanks for this contribution ❤️ Please also extend the docs with the new config option and we are good to go (:

BracketJohn (commented):
Almost there:
- I should've let you know: You have to change the docs of v0.6 as this is the version that will receive this feature. They are located at `docs/v0.6/` - can you move your changes there?
- your changes only effect the `local`-provider - when you move your doc-changes to the `docs/v0.6` part, it should also be clearer, that this is the case

zoey-kaiser (approved):

PR #434, opened 2023-06-10
Fix issue with getToken function and using raw: true in options. 

csprance:
Other.

Next-auth GetTokenParams uses a generic R. It defaults to false. So if you want to use raw: true in the params it will error in typescript. This allows one to pass in the type they want for the raw param. To allow fetching the raw jwt instead of the decoded one.

Checklist:


BracketJohn (approved):
Thanks for this - great contribution ❤️

PR #496, opened 2023-07-25
fix: Don't fetch session during nitro prerender

tesla3327:
During prerendering `getSession` is currently always called, even on non-protected pages, giving this error: `ERROR  [nuxt] [request error] [unhandled] [500] fetch failed (http://localhost:3000/api/auth/session?callbackUrl=http://localhost/)`

This change just checks if we're prerendering or not, using the `x-nitro-prerender` header [set by Nitro during prerendering](https://github.com/unjs/nitro/blob/fddf57e50ed4bd4863a23d80c86645bcea79b884/src/prerender.ts#L155).

I found this issue in `0.5.0` but I'm not sure if you're doing patch releases on this or not.

### DIFF
--- a/src/runtime/plugin.ts
+++ b/src/runtime/plugin.ts
@@ -11,13 +11,13 @@ export default defineNuxtPlugin(async (nuxtApp) => {
   const { getSession } = useAuth()
 
   // Skip auth if we're prerendering
-  let nitroPrerender
+  let nitroPrerender = false
   if (nuxtApp.ssrContext) {
-    nitroPrerender = getHeader(nuxtApp.ssrContext.event, "x-nitro-prerender")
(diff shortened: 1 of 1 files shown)


zoey-kaiser:
Hi @tesla3327 

Thank you for your contribution! You said that you tested and I assume found the issue in 0.5.0. Is this also present in 0.6.0? I would assume so, as your contribution is for v0.6!

genu:
I can confirm that prerender does throw an error in the latest version.

Other than linting, anything else holding this PR back?

andr-ec:
@zoey-kaiser I'm seeing this as well. Can we accept this?

zoey-kaiser:
> Other than linting, anything else holding this PR back?

At the moment, the linting is the only problem!

genu:
@zoey-kaiser Done. Have a look at #521

PR #543, opened 2023-10-11
fix: Do not block the login page in `globalMiddleware`

### LINKED ISSUE #410: Infinite Redirect - Global Middleware
### Environment

"nuxt": "3.4.2",
"@sidebase/nuxt-auth": "0.6.0-beta.2",

### Reproduction

_No response_

### Describe the bug

[Vue Router warn]: Detected a possibly infinite redirection in a navigation guard when going from "/account/settings" to "/auth/signin". Aborting to avoid a Stack Overflow.  Are you always returning a new location within a navigation guard? That would lead to this error. Only return when redirecting or aborting, that should fix this. This might break in production if not fixed.

[nuxt] Calling useRoute within middleware may lead to misleading results. Instead, use the (to, from) arguments passed to the middleware to access the new and old routes.  

### Additional context

_No response_

### Logs

_No response_

zoey-kaiser:
Closes #410 

Ensures the `globalMiddleware` will not block the login page defined in the nuxt.config.

- If blocked, will lead to an endless redirect, between loginPage (gets Blocked) -> loginPage (gets Blocked) causing application to crash
- removes confussion, as the `globalMiddleware` does not block the default authjs login page

Checklist:
- [X] wrote tests / testing not applicable

### DIFF
--- a/docs/content/v0.6/2.configuration/2.nuxt-config.md
+++ b/docs/content/v0.6/2.configuration/2.nuxt-config.md
@@ -132,7 +132,7 @@ type ProviderLocal = {
    */
   endpoints?: {
     /**
-     * What method and path to call to perform the sign-in. This endpoint must return a token that can be used to authenticate subsequent requests.
+     * What method and path to call to perform the sign-in. This endpoint must return a token that can be used to authenticate subsequent requests. This page will also not be blocked by the global middleware.
      *
      * @default { path: '/login', method: 'post' }
      */
--- a/docs/content/v0.6/2.configuration/2.nuxt-config.md
+++ b/docs/content/v0.6/2.configuration/2.nuxt-config.md
@@ -132,7 +132,7 @@ type ProviderLocal = {
    */
   endpoints?: {
     /**
-     * What method and path to call to perform the sign-in. This endpoint must return a token that can be used to authenticate subsequent requests. This page will also not be blocked by the global middleware.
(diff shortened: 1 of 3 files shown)


src/runtime/middleware/auth.ts
```diff
@@ -61,6 +61,14 @@ export default defineNuxtRouteMiddleware((to) => {
     return
   }
 
+  // We do not want to block the login page when the local provider is used
+  if (authConfig.provider?.type === 'local') {
+    const loginRoute: string | null = authConfig.provider?.pages?.login
+    if (loginRoute && loginRoute === to.path) {
```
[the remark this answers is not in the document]
zoey-kaiser: Sure!

docs/content/v0.6/2.configuration/2.nuxt-config.md
```diff
@@ -132,7 +132,7 @@ type ProviderLocal = {
    */
   endpoints?: {
     /**
-     * What method and path to call to perform the sign-in. This endpoint must return a token that can be used to authenticate subsequent requests.
+     * What method and path to call to perform the sign-in. This endpoint must return a token that can be used to authenticate subsequent requests. This page will also not be blocked by the global middleware.
```
BracketJohn, line 135: this is the `endpoints:` specification

zoey-kaiser: I don't think so, it talks about signin and the `endpoints` have a description above it.

Issue #553, opened 2023-10-22
Local provider causes hard refresh when logging in/out/refresh


IsraelOrtuno:
### Describe the bug

Navigation caused by logging in/out/refresh is using `navigateTo(..., { external: true })` since #540 and causes a full page reload whenever any of these actions happen, which is commonly an unwanted behaviour.

As I far as I understand this option was added because just like the PR says, the redirect URL is being constructed using protocols and full paths, so that causes `navigateTo` to complain about it if you don't use `{ external: true }` and maybe even causing errors #429 .

If that's the only reason to add this external parameter we could maybe just try to guess that if `callback` is external somehow. I wonder if there's any downside in just doing something like this:

https://github.com/nuxt/nuxt/blob/88ce9abbdf1252af76b6e7c8b8adf58b7b76589f/packages/nuxt/src/app/composables/router.ts#L134

And maybe making the `external` option optional or even better, automatic.

PR #554, opened 2023-10-23
feat: add set and clear token methods

IsraelOrtuno:
Added `getToken` and `clearToken` to local provider useAuthState composable for convenience. It's quite common to have cases where you have to "log in" the user manually when using this provider like when using social authentication, impersonation from a back-office or a custom register process. This PR just adds a couple of methods to make setting the cookie token more convenient and obvious.

Checklist:
- [X] issue number linked above after pound


zoey-kaiser:
Hi @IsraelOrtuno 👋 

Thank you for your contribution! I think it would be super helpful to create a quick note on this feature in the docs, to explain how to use this improvement. I think it would be good to write a quick section under `docs/nuxt-auth/0.6/resources`.

IsraelOrtuno:
> Hi @IsraelOrtuno 👋
> 
> Thank you for your contribution! I think it would be super helpful to create a quick note on this feature in the docs, to explain how to use this improvement. I think it would be good to write a quick section under `docs/nuxt-auth/0.6/resources`.

Yes I thought so too but wasn't sure where to add it. I have updated the PR with a dedicated docs section which I guess it can be rearranged in the future 👍

zoey-kaiser:
Thanks for adding the docs! I think the texts are good and well written, however after some more thinking, I would place them at the bottom of [this file](https://github.com/sidebase/nuxt-auth/blob/main/docs/content/v0.6/3.application-side/2.session-access-and-management.md) with a new heading, called `useAuthState() Composable`. You could then also mention that this section is meant for use with the local provider. 

Another thing I wanted to point out, AFAIK `useAuthState` is used by both the local and authjs providers, does including this feature break anything or interact with the Authjs provider? If you have not had the chance to test this yet, I can also do it today!

IsraelOrtuno:
I don't think it has anything to do with `authjs` nor breaking anything. This is a `local` provider which is likely to be used with custom backend solutions.

I have also updated the docs with a more extensive information about what `useAuthState` composable does not only for `local` but for `authjs` too 👍

zoey-kaiser (approved):
Thank you so much for your contribution. I will merge this now, wait till I close issue #553 and then move on to releasing this in `0.6.0-rc.1`!

Issue #429, opened 2023-06-06
 Error: Navigating to external URL is not allowed by default. Use `navigateTo (url, { external: true })`


Jeroen-Cox:
### Environment

- Operating System: `Darwin`
- Node Version:     `v19.8.1`
- Nuxt Version:     `3.5.3`
- Nitro Version:    `2.4.1`
- Package Manager:  `yarn@1.22.18`
- Builder:          `vite`
- User Config:      `ssr`, `typescript`, `runtimeConfig`, `modules`, `auth`, `components`, `app`
- Runtime Modules:  `@sidebase/nuxt-auth@0.6.0-beta.2`
- Build Modules:    `-`

### Describe the bug

I setup a basic 'local' authentication using @sidebase/nuxt-auth@0.6.0-beta.2. I can successfully login/logout etc. But everytime nuxt-auth wants to do a redirect I see two errors:

`[Vue warn]: Unhandled error during execution of native event handler`

`Uncaught (in promise) Error: Navigating to external URL is not allowed by default. Use navigateTo (url, { external: true })`

So these errors happen after using signIn() and signOut()

### Additional context

My nuxt.config.ts:

export default defineNuxtConfig({
  // Client side rendering only
  ssr: false,
  // Typescript settings
  typescript: {
    typeCheck: false,
    strict: true,
  },
  runtimeConfig: {
    public: {
      baseURL: process.env.API_BUILDER_URL,
      flowBuilderUrl: process.env.FLOW_BUILDER_URL,
      downloadCloseHost:
        process.env.DOWNLOAD_CLOSE_URL || "https://downloadclose.com",
    },
  },
  modules: ["@sidebase/nuxt-auth"],
  // Auth setup, see https://sidebase.io/nuxt-auth/configuration/nuxt-config
  // TODO: set auth origin for production
  auth: {
    provider: {
      type: "local",
      endpoints: {
        signIn: { path: "/login", method: "post" },
        signOut: { path: "/logout", method: "post" },
        getSession: { path: "/user", method: "get" },
      },
      token: {
        type: "Bearer",
      },
      pages: { login: "/login" },
    },
    baseURL: `${process.env.API_BUILDER_URL}/api`,
    globalAppMiddleware: {
      isEnabled: true,
    },
  },
  components: [
    {
      path: "~/components",
      pathPrefix: false,
    },
  ],
  app: {
    // Page header settings
    head: {
      title: "Close Builder",
      titleTemplate: "%s – Close",
[...]

PR #556, opened 2023-10-23
feat: Added option to set `callbackUrl` as external

### LINKED ISSUE #553: Local provider causes hard refresh when logging in/out/refresh

zoey-kaiser:
Closes #553 

You can now direct to an external url in the signin, signout and session methods:

```ts
signIn({ username, password }, { callbackUrl: 'https://sidebase.io', external: true })
```

Note: If this is set to true, then a hard page refresh will be done, therefore only set `external` to true, if you are really redirecting to a page outside of your application.

Checklist:
- [X] wrote tests / testing not applicable

### DIFF
--- a/docs/content/v0.6/3.application-side/2.session-access-and-management.md
+++ b/docs/content/v0.6/3.application-side/2.session-access-and-management.md
@@ -85,6 +85,9 @@ await signIn(credentials)
 // Trigger a sign-in with a redirect afterwards
 await signIn(credentials, { callbackUrl: '/protected' })
 
+// Trigger a sign-in with a redirect afterwards to an external page (if set, this will cause a hard refresh of the page)
+await signIn(credentials, { callbackUrl: 'https://sidebase.io', external: true })
+
 // Trigger a sign-out
 await signOut()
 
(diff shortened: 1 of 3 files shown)


zoey-kaiser:
CC @IsraelOrtuno and @Jeroen-Cox as this is a follow up and change of how I implemented your solution in #429

IsraelOrtuno:
I think this is good for now! 🎉

PR #566, opened 2023-11-13
docs: Make `0.6.0` docs main documentation

zoey-kaiser:
Makes the `0.6.0` docs the default docs. Does not remove the `0.5.0` docs to ensure we keep these archived. A follow up Pr on https://github.com/sidebase/docs is required and will be linked here.

Edit: Sister PR can be found here: https://github.com/sidebase/docs/pull/149


BracketJohn:
I'm a bit overwhelmed byu these changes: What is happening here? Why do we have to switch around all dirs and cannot just point `latest` to `v0.6` and be happy?

zoey-kaiser:
> I'm a bit overwhelmed byu these changes: What is happening here? Why do we have to switch around all dirs and cannot just point latest to v0.6 and be happy?

The problem is, that we had all the old links (on google, other pages etc.) never point to a version of the docs. If we were to leave them in that folder, we would need to update all the links, wait for google to update etc.

This would make a lot of problems. It would have been nicer if we from the very beginning had versioned paths, but as we did not do this, we now have this issue.

BracketJohn:
I see, but will this be resolved from now on / can we switch to the versioned approach now?

zoey-kaiser:
> I see, but will this be resolved from now on / can we switch to the versioned approach now?

It would either:

- Point all existing links and google searches (for a while until it updates) to a 404 (if we move the old docs into `/v0.5`)
- Open outdated documentation if we keep the order as it is right now.

I don't think this is something we can tackle today.

BracketJohn (approved):

BracketJohn (changes requested):
see above

zoey-kaiser:
> point latest to 0.6

This is something we attempted during the initial versioning PR. The problem I ran into was that I could not create catch all redirects. 
What this essentially means:
- I would need to create a `md` file for every route we used to have in the docs and point each file to the new docs. 

I would have loved to have a catch all route that says: 
- If link has `nuxt-auth/X` point to `nuxt-auth/v0.6/X`, but this is not possible. 

See https://github.com/nuxt-themes/docus/issues/829 for the previous investigation.

BracketJohn:
Ok, and applying this solution without the redirect?

zoey-kaiser:
> Ok, and applying this solution without the redirect?

This would result in many of our old links being inaccessible. This includes google search results until google updates them, which could take a while.

zoey-kaiser:
Added the banner as agreed:


Screenshot 2023-11-13 at 16 02 43

BracketJohn (approved):

BracketJohn:
Nit: It should probably be "You are viewing the docs for the newest version. [...]"

zoey-kaiser:
> Nit: It should probably be "You are viewing the docs for the newest version. [...]"

True, good catch

Screenshot 2023-11-13 at 16 08 14

PR #582, opened 2023-11-22
fix: augment #app rather than more complex subpath

danielroe:
We're seeing some cases where when users augment `#app/nuxt` rather than `#app`, they can lose type support once the (future default) typescript module resolution of 'bundler' is enabled. While searching for repos using that, I also came across this more complex declaration....

This PR shouldn't cause any issues for you but it might prevent issues in future.

related: https://github.com/nuxt/nuxt/issues/24193

### DIFF
--- a/src/runtime/middleware/auth.ts
+++ b/src/runtime/middleware/auth.ts
@@ -22,7 +22,7 @@ type MiddlewareMeta = boolean | {
   navigateUnauthenticatedTo?: string
 }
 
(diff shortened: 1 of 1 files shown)


zoey-kaiser (approved):

zoey-kaiser:
Hi @danielroe 👋 

Thanks for your PR and your continued help recently! This PR actually does resolve some issues people have had with type support before and explains a lot 😆

PR #392, opened 2023-05-14
feat: add a json pointer for the session object

Danielwinkelmann:
Closes #350  .

add a json pointer for the session object.

Checklist:
- [x] issue number linked above after pound (`#`)
- [x] manually checked my feature (playground-local)
- [x] testing not applicable (playground-local)
- [x] screenshot not applicable

PR #592, opened 2023-11-30
[PR changes #392] feat: move pointer and sessionDataType to the SessionConfig

valh1996:
Hi,

As I don't have the rights to push directly on the original PR (#392) by @Danielwinkelmann, I'm opening a new one.

This PR includes all of @Danielwinkelmann's changes and applies @BracketJohn's latest comments.

As a reminder, here is the comment: https://github.com/sidebase/nuxt-auth/pull/392#discussion_r1193562524

Thus `sessionDataType` and the pointer are moved directly into the `SessionConfig` and renamed to avoid redundancy with the "Session" keyword:
```js
session: {
    dataType: { id: 'string' },
    dataResponsePointer: '/data/user'
},
```

Don't hesitate @zoey-kaiser & @BracketJohn  to let me know about any changes you'd like to make so that we can move forward 🙏🏻


zoey-kaiser:
Hi @valh1996 Thanks for taking this over. Could you please resolve the merge conflicts and then ill do a review?

valh1996:
> Hi @valh1996 Thanks for taking this over. Could you please resolve the merge conflicts and then ill do a review?

Hi, it's done, in principle it should be OK. There were many changes to merge, so I've corrected the conflicts manually.

Can you just check that the new `dataType` property in the `SessionConfig` works well with the authjs and refresh providers ? I don't really know them since I only use `local`.

Thanks

zoey-kaiser:
Hi! I have some time on Thursday to do a proper review. If we want to get an alpha version out first it would be great if you could answer the following questions for me, then I don't have to look into the code myself to figure it out 😓 

- What migrations will we expect from people using the `local` provider once we release this?
- This PR allows developers to pharse the sessionData type to allow better type support right?
- Can we also add this to the new `refresh` provider to ensure the two providers are on a similar feature set? (I can also add this afterwards if you don't have time)

valh1996:
No worries, I understand, plus it's an open-source project so I'll wait and thank you again for your work^^

____
> * What migrations will we expect from people using the `local` provider once we release this?

They will have to update their `nuxt.config.ts` as the `sessionDataType` property will no longer exist in the provider config

Instead, they must define the session property as follow:
```ts
session: {
  dataType: {
    id: 'number',
    username: 'string',
  },
  dataResponsePointer: '/data',
  enableRefreshPeriodically: false,
  enableRefreshOnWindowFocus: true,
}
```

Please note the two new properties:
- `dataType` they will have to move their old `sessionDataType` configuration to this one
- `dataResponsePointer` the pointer


> This PR allows developers to pharse the sessionData type to allow better type support right?

Yes, for example if an API returns something like :
```json
{
  "data": {
    "id": 1,
    "username": "val"
  },
  "success": false
}
```
So with the configuration above, and in particular the pointer, my user can be returned directly, so I don't have to do `data.value.data` unnecessarily, but directly `data.value`

> Can we also add this to the new refresh provider to ensure the two providers are on a similar feature set? (I can also add this afterwards if you don't have time)

Since the configuration is in the session and no longer in the provider itself, this should normally work for all providers. I will check this and correct if necessary.

valh1996:
Hey @zoey-kaiser !
I took a little time this morning to put things in order:

1. The `enableRefreshPeriodically` and `enableRefreshOnWindowFocus` params are now optional in the `nuxt.config.ts` session object. Default values are used.

2. I've taken over some of the changes @Danielwinkelmann made to the `jsonPointerGet` method to have better typings using generics. **There is NO breaking changes.**

3. I've therefore specified the return type wherever the `jsonPointerGet` method is used, even in the `refresh` provider. 

4. I've fixed a bug I left behind by copying Daniel's changes.

___

Since point 1 has been done, users will only have to remove the `sessionDataType` from the `provider` in the `nuxt.config.ts` and add the two new properties to the `session` object: 
```ts
session: {
  dataType: {
    id: 'number',
    username: 'string',
  },
  dataResponsePointer: '/data',
  // enableRefreshPeriodically: false,   -> now optional
  // enableRefreshOnWindowFocus: true    -> now optional
}
```

Moreover, since the refresh provider is based on the local provider's `getSession()`, it works out-of-the-box.

zoey-kaiser:
Hi @valh1996 

Thank you so much for continuing to push this! I really like the direction this PR is going in. There is one more type issue, but I think we are getting close to being able to merge 🥳

Therefore I wanted to quickly give you an outlook on the next steps:

- As this is a breaking change where the config needs to be updated, I will need to do a minor release for this (instead of a patch). 
- We currently have an Alpha patch released with the new refresh provider. On Thursday I will therefore do a comprehensive test of this and if everything goes well, release the patch version. 
- Once this is done, I can merge this PR and make an alpha release of the next minor update! 

Due to this, I will also need to check how I want to sort the docs, to ensure that there is a clear separation between the nuxt config setup from version `0.6` to the new `0.7` we would be releasing with this change. 

I will discuss this once more with my team internally to see how we want to best approach this! I will keep you updated.

zoey-kaiser (changes requested):
One issue that arose from my first review is the moving of `sessionDataType` from the providers to the `SessionConfig`.

The reason why `sessionDataType` was not in inside the `SessionConfig` before was that it is not compatible with the `authjs` provider. We generally set the standard that:

Any config that is not applicable to all providers will not be in the general config options, but kept inside of the provider specific config. 

Can we therefore move the `sessionDataType` back to the provider specific config? This would also avoid users having to change their config with this change!

valh1996:
> One issue that arose from my first review is the moving of `sessionDataType` from the providers to the `SessionConfig`.
> 
> The reason why `sessionDataType` was not in inside the `SessionConfig` before was that it is not compatible with the `authjs` provider. We generally set the standard that:
> 
> Any config that is not applicable to all providers will not be in the general config options, but kept inside of the provider specific config.
> 
> Can we therefore move the `sessionDataType` back to the provider specific config? This would also avoid users having to change their config with this change!

I understand, but in this case it doesn't make sense to put the new `sessionDataResponsePointer` in the `SessionConfig` either then?

And if neither of them belong in SessionConfig, I don't understand why changes were requested on the original PR #392 ?

zoey-kaiser:
> And if neither of them belong in SessionConfig, I don't understand why changes were requested on the original PR https://github.com/sidebase/nuxt-auth/pull/392 ?

AFAIK the original comment, proposed that it should be moved into the session config and also added to the authjs provider. As cool as this would be, I can understand if this does not make it into this PR. This is why I am recommending to either:

- add support for authjs
- move the session data type back to the providers (local and refresh)

valh1996:
Hey @zoey-kaiser I apologize for the delay, I'm very busy ATM.

> add support for authjs

I've never used this provider, so if you'd like to lend a hand, I'd love to, otherwise I'll take a look, but not before the beginning of next year.

zoey-kaiser:
> I apologize for the delay, I'm very busy ATM.

No problem at all! We are super thankful for your contribution ❤️ 

> I've never used this provider, so if you'd like to lend a hand, I'd love to, otherwise I'll take a look, but not before the beginning of next year.

That is no problem at all. For now I will forward to this @BracketJohn, to have a look at your current implementation and also discuss the next steps for adding authjs support. This could probably also be done by us in a follow up! 
I expect him to get back to you in the next 2 weeks!

valh1996:
Hi @zoey-kaiser,

Thanks for your reply, I took the time to move back the session data type back to the providers (local and refresh) as requested. So we can certainly go ahead and implement this for the authjs provider at a later date.

I've wrapped both properties in a session object for greater clarity, so we get the best of both worlds, what do you think?

zoey-kaiser (approved):

zoey-kaiser:
Hi @valh1996, quick update from my side 👋 

Sorry its been taking a while. I wanted @BracketJohn to run over this PR on last time, as he made the initial proposes in the original PR. Therefore I have reached out to him and would want to wait for his approval before merging. Sorry for the wait!

valh1996:
Hey @zoey-kaiser, any update please?

phoenix-ru:
Hi @valh1996, could you please resolve the last conflict? Your PR is on our shortlist, so we want to push it

valh1996:
Hi @phoenix-ru
I've updated the PR on the main branch, which should correct the problem. Let me know

phoenix-ru:
@valh1996 Thanks for a quick reaction. We have your PR planned for 0.7 because it includes some breaking changes as you already know.
For that we need to ensure stability of 0.6 first. This means adding E2E tests to our CI.
Sorry for delays, we just need some infra setup first. Will try to iterate fast.

playground-local/nuxt.config.ts
```diff
@@ -15,12 +15,14 @@ export default defineNuxtConfig({
       token: {
         signInResponseTokenPointer: '/token/accessToken'
       },
-      sessionDataType: { id: 'string', email: 'string', name: 'string', role: 'admin | guest | account', subscriptions: "{ id: number, status: 'ACTIVE' | 'INACTIVE' }[]" }
+      session: {
+        dataType: { id: 'string', email: 'string', name: 'string', role: 'admin | guest | account', subscriptions: "{ id: number, status: 'ACTIVE' | 'INACTIVE' }[]" },
+        dataResponsePointer: '/data/user'
```
phoenix-ru, line 20: @valh1996 Interestingly enough, the tests caught some issue with your PR. I believe that this configuration should be just
```suggestion
        dataResponsePointer: '/'
```

phoenix-ru (changes requested):

playground-local/nuxt.config.ts
phoenix-ru: Confirmed. Running `pnpm test:e2e` inside `playground-local` with this configuration fixed, all the tests pass.

valh1996: @phoenix-ru Well done! An oversight on my part indeed

phoenix-ru:
@valh1996 Amazing, I'll consult with @zoey-kaiser if we can merge the PR and do a new release (0.8.0) or if there are any other non-breaking changes which can be included in 0.7

valh1996:
Hello @phoenix-ru any news on this one please?

phoenix-ru:
Hi @valh1996 , sorry for keeping you waiting.
I don't think I can follow up this week, but I will put this to high priority for the next week.
I think our 0.7 is good already, and we already have several breaking changes to prepare 0.8, so I would try to merge this asap (not guaranteeing next week, but it's possible).

In the meantime, could you please resolve merge conflicts?

phoenix-ru (approved):

valh1996:
sorry @phoenix-ru, I really haven't had time to look at the conflicts, no time at the moment, thanks for doing so !

phoenix-ru:
No problem @valh1996 , I have resolved them. Sorry for keeping you waiting and good work!

valh1996:
Hi @phoenix-ru! I've finally got around to testing the latest release, which contains this PR.

Unfortunately, I can't get it to work. Have you had any feedback along these lines?

This is my configuration, which basically worked in my tests during PR development:
```ts
auth: {
  baseURL: `${API_BASE_URL}/v1/auth`,
  globalAppMiddleware: true,
  provider: {
    type: 'local',
    endpoints: {
      signIn: { path: '/login', method: 'post' },
      getSession: { path: '/me', method: 'get' },
      signOut: false
    },
    token: {
      signInResponseTokenPointer: '/data/token',
    },
    session: {
      dataType: { id: 'number', username: 'string' },
      dataResponsePointer: '/data'
    }
  },
},
```

When I sign in, I have this [error](https://github.com/sidebase/nuxt-auth/blob/main/src/runtime/helpers.ts#L69):
```
Error: Invalid reference token: data
```

Here is the login data :
```json
{
  "data": {
    "token": "xxx",
    "expire_in_ms": 900000
  },
  "success": true,
  "status": 200
}
```
Am I missing something obvious or could there be a bug please?

Issue #596, opened 2023-12-03
How to mock useAuth in Vitest ? 


ErinTran15:
### Describe the bug

I've been trying to do vi.mock and unplugin-auto-import to mock it but none of them works
when I config it in unplugin-auto-import in vitest.config.ts like this
```
 imports: [
      'vue',
      'vue-router',
      {
        '@sidebase/nuxt-auth: [
          "useAuth"
        ],
      },
    ],
```
surely it won't work because its the wrong path
but when I import it with the correct path like in auto import path it will give me error: 
```
 imports: [
      'vue',
      'vue-router',
      {
        '@sidebase/nuxt-auth/dist/runtime/composables/authjs/useAuth': [
          "useAuth"
        ],
      },
    ],
```
Missing "./dist/runtime/composables/authjs/useAuth" specifier in "@sidebase/nuxt-auth" package

pls is there any other ways I could mock it ? Really appreciate if you guys can help me

zoey-kaiser:
Hi! Mocking is generally a bit harder to accomplish inside of the module, as everyone has different session data. The way we handled it internally is as follows:

1.) Inside of `/tests/mock/setup.ts` we created a new "mocked" version of `nuxt-auth`:

```ts
import { createResolver, defineNuxtModule } from '@nuxt/kit'

export default defineNuxtModule({
  setup: (_options, nuxt) => {
    const { resolve } = createResolver(import.meta.url)
    const pathToMocks = resolve('./nuxt-auth.ts')

    nuxt.hook('imports:extend', (_imports) => {
      _imports.push({ name: 'useAuth', from: pathToMocks })
    })

    nuxt.hook('nitro:config', (nitroConfig) => {
      if (!nitroConfig.alias) {
        throw new Error('Alias must exist at this point, otherwise server-side cannot be mocked')
      }
      nitroConfig.alias['#auth'] = pathToMocks
    })
  },
})
```

2. Inside of this file we referenced a mocked version of the composable used inside of NuxtAuth written in `/tests/mock/nuxt-auth.ts`

```ts
import { eventHandler } from 'h3'

export const MOCKED_USER = { user: { role: 'admin', email: 'test@example.com', name: 'John Doe' } }

// App-side mocks
export const useAuth = () => ({
  data: ref(MOCKED_USER),
  status: ref('authenticated'),
  getSession: () => MOCKED_USER,
  signOut: () => {},
})

// Server-side mocks
export const getServerSession = () => MOCKED_USER
[...]

Issue #552, opened 2023-10-18
Handle disabled state gracefully


peterbud:
### Describe the feature

nuxt-auth is working great in my project. However I'm using also `histoire` to build and test my components in isolation. In that case I set `isEnabled: false` for nuxt-auth in nuxt.config.ts

The problem is that in that case I get a bunch of errors everywhere where on the server side I have #auth import:
`Could not resolve import "#auth"`

Question: would it be possible to handle such cases more gracefully and avoid those errors, when the module is disabled?

peterbud:
I see similar issue has raised #454

zoey-kaiser:
Hi @peterbud!

So we do a bunch of testing in our projects as well. The way we approached it was to override the `useAuth` composable with our custom custom one made for testing. 

This composable then also mocks a user account and returns demo functions that our testing suite can use. 

You can add this into a testing file:

```ts
const pathToMocks = resolve('./nuxt-auth.ts')

nuxt.hook('imports:extend', (_imports) => {
  _imports.push({ name: 'useAuth', from: pathToMocks })
})
```

This will then overwrite the `useAuth` composable with the mocked one, allowing you to use it in testing.

peterbud:
Thanks for the reply and the help. Can you point me to an example where this is used somewhere?

When I have tried to add this hook to `nuxt.config.ts`, I got the following error:
```
Duplicated imports "useAuth", the one from "[...]mock-useAuth.ts" has been ignored and "[....]/node_modules/.pnpm/@sidebase+nuxt-auth@0.6.0-rc.1_next-auth@4.22.5_vue@3.3.7/node_modules/@sidebase/nuxt-auth/dist/runtime/composables/authjs/useAuth" is used
```

zoey-kaiser:
> Can you point me to an example where this is used somewhere?

This is very dependent on your testing setup. I do not have much experience with historie, but I assume there is a setup file, similarly to how storybook does this. Inside this setup file you can then inject this new hook. You have to ensure it is only being used for your historie setup and does not override the actual composable in the app.

Issue #454, opened 2023-06-22
could not resolve import when module isEnabled = false


agracia-foticos:
### Environment

- Operating System: `Linux`
- Node Version:     `v18.16.0`
- Nuxt Version:     `3.5.3`
- Nitro Version:    `2.4.1`
- Package Manager:  `npm@9.5.1`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

We are using the same code, for multi-site APP... same sites has login and others sites hasnt login.

We have a wrapper to use useAuth, but we have a problem with 
src/server/routes/api/auth/[...].ts file...

we have the module disabled but when  in auth file try to import import { NuxtAuthHandler } from '#auth' throws an error

`(node-resolve plugin) Could not resolve import "#auth" in /home/foticos/var/www/wanapix_next/src/server/routes/api/auth/[...].ts using imports defined in /home/foticos/var/www/wanapix_next/package.json.
(node-resolve plugin) Could not resolve import "#auth" in /home/foticos/var/www/wanapix_next/src/server/routes/api/auth/[...].ts using imports defined in /home/foticos/var/www/wanapix_next/package.json.
"#auth" is imported by "src/server/routes/api/auth/[...].ts", but could not be resolved – treating it as an external dependency.`

If we use visual studio debugger, we cant start the app, only can with npx nuxi build and node start

### Describe the bug

We are using the same code, for multi-site APP... same sites has login and others sites hasnt login.

We have a wrapper to use useAuth, but we have a problem with 
src/server/routes/api/auth/[...].ts file...

we have the module disabled but when  in auth file try to import import { NuxtAuthHandler } from '#auth' throws an error

`(node-resolve plugin) Could not resolve import "#auth" in /home/foticos/var/www/wanapix_next/src/server/routes/api/auth/[...].ts using imports defined in /home/foticos/var/www/wanapix_next/package.json.
(node-resolve plugin) Could not resolve import "#auth" in /home/foticos/var/www/wanapix_next/src/server/routes/api/auth/[...].ts using imports defined in /home/foticos/var/www/wanapix_next/package.json.
"#auth" is imported by "src/server/routes/api/auth/[...].ts", but could not be resolved – treating it as an external dependency.`

If we use visual studio debugger, we cant start the app, only can with npx nuxi build and node start

agracia-foticos:
any solution?

agracia-foticos:
[...]

PR #611, opened 2023-12-15
feat(imports): Add minimal stubs useAuth and useAuthState imports

peterbud:
When `nuxt-auth` module is disabled (`isEnabled=false`) all `#auth` imports were failing. For mocking and testing this presented a situation which was impossible to handle properly. This commit adds a stub for these imports on the client side in case the `nuxt-auth` module is disabled, opening a way to mock the modules properly without worrying the missing imports.

Contributes to #454 #552 #596.

Checklist:
- [X] testing not applicable


zoey-kaiser:
Hi @peterbud 👋 

Thank you for your contribution! I think this is a really nice change, especially for testing suits! Do you think we may want to go one additional step and allow mocking of the composables? This way you could integrate e2e tests in a nicer way!

peterbud:
I think mocking would be possible, but I'm not sure what additional step you mean. Could you pls clarify?

zoey-kaiser:
> I think mocking would be possible, but I'm not sure what additional step you mean. Could you pls clarify?

I wrote a lengthy comment on how we handle mocking stubs inside of our internal applications here: https://github.com/sidebase/nuxt-auth/issues/596#issuecomment-1864222564

I also quickly touched on what would need to be done, to properly add mocking into our module.

peterbud:
Thanks for the great summary at the other issue. I think that mocking setup would be best covered by adding it to the playground/example. In this PR I don't see it feasible to add a mocking infrastructure, as this PR is focusing on ensuring that there is _some_ stubs to import when the module is not enabled.

zoey-kaiser:
Hi @peterbud 👋 

> Thanks for the great summary at the other issue. I think that mocking setup would be best covered by adding it to the playground/example. In this PR I don't see it feasible to add a mocking infrastructure, as this PR is focusing on ensuring that there is some stubs to import when the module is not enabled.

Sounds good!

I wanted to let you know that I will upstream this PR to another colleague. Code wise I approve the changes, however I wanted to check with him, if this is how he would handle the `isEnabled` property. I expect him to get around in the next 2 weeks. Thank you for you patience!

phoenix-ru:
@peterbud Could you please fix TS errors (simply add `type` modifier to imports)

peterbud:
Done

src/module.ts
```diff
@@ -155,18 +154,32 @@ export default defineNuxtModule<ModuleOptions>({
...
+        {
+          name: 'useAuth',
+          from: resolve('./runtime/composables/stub/useAuth')
+        },
+        {
+          name: 'useAuthState',
+          from: resolve('./runtime/composables/stub/useAuthState')
+        }
+      ])
+    }
```
phoenix-ru, line 182: I believe you need to do a
```ts
if (!options.isEnabled) {
  return
}
```

before step 6?

Otherwise you end up adding runtime plugins and middleware, is this intended?

phoenix-ru:
@peterbud Could you please elaborate why this PR was closed?

peterbud:
Due to lack of progress. For 4 months there was no traction with this PR. I understand if maintainers are busy.

PR #615, opened 2023-12-18
feat: local provider prevent signIn flow

### LINKED ISSUE #614: Local provider: prevent signIn flow returned by signUp method
### Describe the feature

It could be useful to prevent the returned signIn flow inside the signUp method.
```
source: src/runtime/composables/local/useAuth.ts

const signUp = async (credentials: Credentials, signInOptions?: SecondarySignInOptions) => {
  const nuxt = useNuxtApp()

  const { path, method } = useTypedBackendConfig(useRuntimeConfig(), 'local').endpoints.signUp
  await _fetch(nuxt, path, {
    method,
    body: credentials
  })
  
  // boolean flag

  return signIn(credentials, signInOptions)
}
```

### Additional information

_No response_

vettndr:
Closes #614.

I'm glad to discuss further improvements.
@zoey-kaiser

### DIFF
--- a/src/runtime/composables/local/useAuth.ts
+++ b/src/runtime/composables/local/useAuth.ts
@@ -109,10 +109,14 @@ const signUp = async (credentials: Credentials, signInOptions?: SecondarySignInO
   const nuxt = useNuxtApp()
 
   const { path, method } = useTypedBackendConfig(useRuntimeConfig(), 'local').endpoints.signUp
-  await _fetch(nuxt, path, {
+  const response = await _fetch<Record<string, any>>(nuxt, path, {
     method,
     body: credentials
   })
+
+  if(signInOptions.stopLogin) {
+    return response
+  }
 
   return signIn(credentials, signInOptions)
 }
--- a/src/runtime/composables/local/useAuth.ts
+++ b/src/runtime/composables/local/useAuth.ts
@@ -1,6 +1,6 @@
 import { readonly, Ref } from 'vue'
 import { callWithNuxt } from '#app/nuxt'
-import { CommonUseAuthReturn, SignOutFunc, SignInFunc, GetSessionFunc, SecondarySignInOptions, RegisterOptions } from '../../types'
+import { CommonUseAuthReturn, SignOutFunc, SignInFunc, GetSessionFunc, SecondarySignInOptions, SignUpOptions } from '../../types'
 import { _fetch } from '../../utils/fetch'
 import { jsonPointerGet, useTypedBackendConfig } from '../../helpers'
 import { getRequestURLWN } from '../../utils/callWithNuxt'
@@ -105,7 +105,7 @@ const getSession: GetSessionFunc<SessionData | null | void> = async (getSessionO
   return data.value
 }
 
-const signUp = async (credentials: Credentials, signInOptions?: SecondarySignInOptions, registerOptions?: RegisterOptions) => {
+const signUp = async (credentials: Credentials, signInOptions?: SecondarySignInOptions, signUpOptions?: SignUpOptions) => {
   const nuxt = useNuxtApp()
 
   const { path, method } = useTypedBackendConfig(useRuntimeConfig(), 'local').endpoints.signUp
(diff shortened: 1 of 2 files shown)


src/runtime/types.ts
```diff
@@ -395,6 +395,11 @@ export interface SecondarySignInOptions extends Record<string, unknown> {
    * @default false
    */
   external?: boolean;
+  /** Prevent the signIn flow during registration and return the signUp response
+   *
+   * @default false
+   */
+  stopLogin?: boolean;
```
zoey-kaiser, line 402: This is something we only want to use for the registration flow, but adding it here, would also allow you to pass it for the signin flow. I think we need to find a way to:

- Not modify the signin options (as for registration these are then passed onto the `signIn` function
- We don't want to change the order of props that can be passed (to not make breaking changes)

I think we need to either: 
- Expand the `signInOptions` type by creating a `registerOptions` type with the added value
- pass the value as a third optional parameter

Also lets rename it to `preventLoginFlow`

src/runtime/composables/local/useAuth.ts
```diff
@@ -109,11 +109,15 @@ const signUp = async (credentials: Credentials, signInOptions?: SecondarySignInO
...
 
   const { path, method } = useTypedBackendConfig(useRuntimeConfig(), 'local').endpoints.signUp
-  await _fetch(nuxt, path, {
+  const response = await _fetch<Record<string, any>>(nuxt, path, {
     method,
     body: credentials
   })
 
+  if(signInOptions?.stopLogin) {
+    return response
```
zoey-kaiser, line 118: I would not return different values based on if you want the signIn flow to happen or not, as I feel like this would lead to confusion.

zoey-kaiser (changes requested):

vettndr:
@zoey-kaiser 
thank you for the feedbacks.
I've just pushed some improvements as you suggested me. 🚀

src/runtime/composables/local/useAuth.ts
```diff
@@ -105,15 +105,19 @@ const getSession: GetSessionFunc<SessionData | null | void> = async (getSessionO
   return data.value
 }
 
-const signUp = async (credentials: Credentials, signInOptions?: SecondarySignInOptions) => {
+const signUp = async (credentials: Credentials, signInOptions?: SecondarySignInOptions, registerOptions?: RegisterOptions) => {
   const nuxt = useNuxtApp()
 
   const { path, method } = useTypedBackendConfig(useRuntimeConfig(), 'local').endpoints.signUp
-  await _fetch(nuxt, path, {
+  await _fetch<Record<string, any>>(nuxt, path, {
```
zoey-kaiser, line 112: Why do we add types to `_fetch` here? Generally speaking I am not a fan of adding any type with `any` as it is pretty ambiguous.

zoey-kaiser (changes requested):
Hi! Thanks for addressing my first review, I hope everything made sense 😊

src/runtime/composables/local/useAuth.ts
vettndr: Yeah, I am not fan too.
It's a refuse coming from the previous changes, I've added just like in the n.19 line.
I'm going to delete it 🚀

src/runtime/composables/local/useAuth.ts
```diff
@@ -105,7 +105,7 @@ const getSession: GetSessionFunc<SessionData | null | void> = async (getSessionO
   return data.value
 }
 
-const signUp = async (credentials: Credentials, signInOptions?: SecondarySignInOptions) => {
+const signUp = async (credentials: Credentials, signInOptions?: SecondarySignInOptions, registerOptions?: RegisterOptions) => {
```
zoey-kaiser, line 108: Sorry to bug you one last time, but lets rename all of the options from `registerOptions` to `signUpOptions` to be more inline with the function name and then we are good to go 🥳

zoey-kaiser (changes requested):

zoey-kaiser (approved):

root5427:
Hey folks! Huge thanks to @vettndr  and @zoey-kaiser for their fantastic contributions! 🌟 Any updates on when these changes will be merged? 🚀

zoey-kaiser:
Hi @root5427 👋 

We generally do our updates on Mondays! You can now use this in version https://github.com/sidebase/nuxt-auth/releases/tag/0.6.4!

vettndr:
@zoey-kaiser 
Yeah! I'm so happy for this 🥳

PR #637, opened 2024-01-13
Fix: Update useAuth.ts

dennismwas:
This pull request fixes an issue with the sign in options when calling the Sign in Method, the primaryOptions when set to 'credentials' the sigin payload results in the string being split to single characters.
Thiis fix adresses that issue.

Example showcase

```js
  const res = await signIn('credentials', {
        email: 'mwas@mail.com',
        password: 'qwerty'
      });
```

image

### DIFF
--- a/src/runtime/composables/local/useAuth.ts
+++ b/src/runtime/composables/local/useAuth.ts
@@ -19,7 +19,7 @@ const signIn: SignInFunc<Credentials, any> = async (credentials, signInOptions, 
   const response = await _fetch<Record<string, any>>(nuxt, path, {
     method,
     body: {
-      ...credentials,
+      credentials:credentials,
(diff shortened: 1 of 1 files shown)


zoey-kaiser:
Hi @dennismwas 👋 

Thank you for taking the time to look into this! However, I don't believe this is an issue with the code. The `local` and `refresh` providers do not take a first parameter, as you describe above.

```ts
signIn({
	email: 'mwas@mail.com',
	password: 'qwerty'
})
```

The `credentials` parameters refers to the object you are supposed to pass, not a string. The second object are the further options you would like to pass. Therefore if you were to "unspread" the first one and then passed a `credentials` object, these would not be passed to the `fetch` request correctly, therefore requiring developers to always pass a string first, which we would not even read here.

PR #639, opened 2024-01-17
fix: move vue server plugin out of `server/` directory

danielroe:
We currently apply some protections to help users avoid importing nitro in the vue part of the app and vice versa.

With https://github.com/nuxt/nuxt/pull/25162 we expanded the test cases (and now test for `server\/(api|routes|middleware|plugins)\/`).

We might relax this regexp (cc: @pi0) as it wasn't originally intended to catch module usage (it might be helpful to keep anyway!) but I spotted that your Nuxt server plugin was in the `server/` directory when I think it would more helpfully belong in the `plugins/` directory, mirroring a normal Nuxt project.

Checklist:

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -227,7 +227,7 @@ export default defineNuxtModule<ModuleOptions>({
 
     // 7.2 Add a server-plugin to refresh the token on production-startup
     if (selectedProvider === 'refresh') {
-      addPlugin(resolve('./runtime/server/plugins/refresh-token.server'))
+      addPlugin(resolve('./runtime/plugins/refresh-token.server'))
     }
 
     logger.success('`nuxt-auth` setup done')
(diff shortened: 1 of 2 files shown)


pi0:
We might relax some rules to "warning only" for import protections also reducing chance of breaking Nuxt users by upstream change. but definitely makes sense to refactor this way 👍🏼

BracketJohn (approved):
LGTM & Thanks you two - very nice to proactively apply those changes here (:

flrnt:
with this fix, it is now possible to build the app without crashing it or silently failed it once in production. A quick patch release would be welcome @zoey-kaiser 😍

zoey-kaiser:
> with this fix, it is now possible to build the app without crashing it or silently failed it once in production. A quick patch release would be welcome @zoey-kaiser 😍

Released in https://github.com/sidebase/nuxt-auth/releases/tag/0.6.5! Sorry for the wait, I was coming home from a trip ❤️

PR #641, opened 2024-01-19
docs: update local and refresh providers guide

### LINKED ISSUE #606: Custom login body
### Describe the feature

The documentation states that the `login` endpoint for local & refresh providers must match a `username` & `password` schema
Screenshot 2023-12-13 150111

We're using the same auth endpoint for Android and iOS apps. Enforcing the backend, Android, and iOS implementation to match the one described above is a bit of a hassle (It's merely rewording but they'll be doing that across the codebases). We're also using phone number instead of a username, so sending that value in a username field is a bit counter-intuitive.

It would be great if we can customize the path of the credentials. Maybe with the same approach as https://github.com/sidebase/nuxt-auth/pull/592?

Thanks a lot for your work and the great module 🙏

### Additional information

_No response_

MuhammadM1998:
Resolves #606.

This updates the `local` & `refresh` providers docs to indicate that we can pass any payload to the `signIn` method, not just a `user` & `password`. I've also added an example and reworded some parts that I felt was not clear enough. Feel free to reedit if needed 🙏


zoey-kaiser:
Hi @MuhammadM1998 👋 

Thanks for updating the docs so thoroughly! I am doing a release at the moment, which is why I quickly wanted to chim in here. I will look at these docs in the coming days. As your docs can update, whichout doing a release, ill prio this after the PRs with code changes!

MuhammadM1998:
Nice to know thanks! Waiting for the release and the new session pointer 👀

zoey-kaiser (dismissed):

zoey-kaiser:
Hi @MuhammadM1998 👋 

Please resolve the merge conflicts and then we can merge this! AFAIK This will also fix an issue in our changes:

The backend response returns an object named `token` not a single value. See the references to this here: https://github.com/sidebase/nuxt-auth/pull/649

MuhammadM1998:
@zoey-kaiser The conflict is fixed now. Feel free to quickly rereview

docs/content/1.getting-started/3.quick-start.md
```diff
@@ -60,69 +60,148 @@ export default NuxtAuthHandler({
...
-and return a token that can be used to authenticate future requests in the response body, e.g., like:
+If the backend response included a `token` key, the user will be signed successfully.
+
 ```ts
+// Backend response
 {
-    token: {
-        accessToken: 'eyBlaBlub'
-    }
+  token: 'eyBlaBlub',
```
zoey-kaiser, line 157: In all of our examples (including the linked branch on the example) we return `token` as an object and not a string. 

https://github.com/sidebase/nuxt-auth/blob/main/playground-local/server/api/auth/login.post.ts#L30
https://github.com/sidebase/nuxt-auth-example/blob/add-local-provider/server/api/auth/login.post.ts#L30

I think it would make sense to mirror this in our docs, as this is how it appears in all examples!

zoey-kaiser (changes requested):
One minor nit 😄

docs/content/1.getting-started/3.quick-start.md
MuhammadM1998: I agree. Will push a change

zoey-kaiser (approved):

zoey-kaiser:
Amazing! Thank you so much for the changes. I think you really improved the docs for the refresh provider. I apologize for it taking longer to merge this PR, the start of the year was pretty crazy!

Issue #650, opened 2024-01-26
Invalid Cookie name


carbontwelve:
### Describe the bug

The cookie name being set as `auth:token` includes the separator `:` which is **invalid** as a cookie name.

[RFC 6256](https://www.rfc-editor.org/rfc/rfc6265#section-4.1.1) defines the cookie name as being a token:
> <token, defined in [[RFC2616], Section 2.2](https://www.rfc-editor.org/rfc/rfc2616#section-2.2)>

With RFC2616 stating that a token is `1*<any CHAR except CTLs or separators>`.

The remedy here would be to replace the usage of `:` with `_` which is an allowed character.

### Additional context

This problem came up when I was deploying a Nuxt project to aws using the aws cdk when creating a `CachePolicy` configured to pass through the `auth:token` cookie aws refused to complete citing the cookie name containing invalid characters.

PR #652, opened 2024-02-01
feat: added token prefix option

FAL-coffee:
We will make it possible to change the prefix of the token stored in the Cookie upon refresh when using the local type or refresh type. 

The reason this is necessary: When multiple environments are launched on localhost, conflicts occur in the token storage locations.

Checklist:


carbontwelve:
May I suggest a slight modification, rather than defining the prefix, make both the token cookie and refresh cookie names configurable doing so would then solve for #650 due to `:` not being a valid character in a cookie name.

FAL-coffee:
@carbontwelve 

that's a perspective I hadn't considered. Thank you for sharing it! 
I'll make the revisions and commit them soon.

FAL-coffee:
I have created the following pull request: https://github.com/sidebase/nuxt-auth/pull/654. I will close this PR now.

PR #654, opened 2024-02-02
feat: added token, refreshToken cookieName option

### LINKED ISSUE #650: Invalid Cookie name

FAL-coffee:
Closes #650 .

---
Allow the user to change the property name where tokens are stored in the cookie when using local or refresh types.

---
**Why this is needed:**
 Conflicts occur with where tokens are stored when multiple environments are launched on localhost.
It also solves the problem that issue #650 refers to.

---
Checklist:


phoenix-ru (approved):

phoenix-ru:
Looks good to me.
@zoey-kaiser OP is right in #650, should we use `auth.token` and `auth.refresh-token` as default cookie names instead?

FAL-coffee:
> Looks good to me. @zoey-kaiser OP is right in #650, should we use `auth.token` and `auth.refresh-token` as default cookie names instead?

In my personal opinion, changing the default CookieName could pose a risk of logging users out across all existing services when updating the library version. 
Additionally, since restrictions only apply in certain cases, I believe it is unnecessary to change the default CookieName.

docs/content/2.configuration/2.nuxt-config.md
```diff
@@ -193,6 +193,13 @@ type ProviderLocal = {
      * @example Beer
      */
     type?: string,
+    /**
+     * It refers to the name of the property when it is stored in a cookie.
+     *
+     * @default auth:token
```
phoenix-ru, line 199: Please, change `auth:token` to `auth.token` everywhere

src/module.ts
```diff
@@ -75,12 +76,14 @@ const defaultsByBackend: {
...
       signInResponseTokenPointer: '/token',
       type: 'Bearer',
+      cookieName: 'auth:token',
       headerName: 'Authorization',
       maxAgeInSeconds: 5 * 60,
       sameSiteAttribute: 'none' // 5 minutes
     },
     refreshToken: {
       signInResponseRefreshTokenPointer: '/refreshToken',
+      cookieName: 'auth:refresh-token',
```
phoenix-ru, line 86: Do the same for `auth:refresh-token`

phoenix-ru (changes requested):

phoenix-ru:
@FAL-coffee I have functionally tested your changes. Please, address the review and it can be safely merged.

FAL-coffee:
@phoenix-ru 
I have committed the change to switch the default delimiter from ':' to '.'!

phoenix-ru (approved):

PR #665, opened 2024-02-15
Bump dependencies and add `ModuleOptionsNormalized` type

phoenix-ru:
Addresses known security advisories. Strengthens the type safety of the module.

Closes #660

PR #667, opened 2024-02-16
fix (refresh-token): handle error in refresh-token plugin and fix reshOnlyToken type

anjarupnik:
While implementing refresh token I ran into issue that whole app was crashing when there was an error in refreshToken call. This was due to not handling errors in refresh-token.server.ts, so I opened this PR with proposition to wrap this call in try catch block and reset token and refresh token in case of error.
Also I noticed that refreshOnlyToken type was not correct (it should be boolean that can be set to false, since it is set to true by default in defaultsByBackend).

Checklist:


phoenix-ru:
@anjarupnik Could you please merge the latest `main` to your branch? I did some type improvements in https://github.com/sidebase/nuxt-auth/pull/665 and it also touched the code which you suggest wrapping in try-catch.

phoenix-ru (approved):

phoenix-ru:
@zoey-kaiser Could you do a functional review and merge if all good?

PR #697, opened 2024-03-01
docs: Add missing middleware option for guestmode

anischihi:
### ❓ Type of change

### 📚 Description
I tried the guest middleware, it did not work (n redirect, no unauthenticatedOnly behaviour). Added the missing snippet, everything works like a charm. I think has been missed in the docs. You can try for yourself.

### 📝 Checklist

### DIFF
--- a/docs/content/3.application-side/5.guest-mode.md
+++ b/docs/content/3.application-side/5.guest-mode.md
@@ -15,6 +15,7 @@ This behavior is useful for login pages that you don't want to be visitable by l
 Briefly summarized, you can enable guest mode on a page by passing the following configuration:
 ```ts
 definePageMeta({
+  middleware: 'auth',
   auth: {
     unauthenticatedOnly: true,
     navigateAuthenticatedTo: '/profile',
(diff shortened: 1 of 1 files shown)


MuhammadM1998:
I suggested this in an issue before but Zoey said this is expected when turning off the global middleware (You don't need to add this line if global middleware is enabled). Maybe a hint that this line is needed _only_ when disabling global middleware would be best?

zoey-kaiser:
> I suggested this in an issue before but Zoey said this is expected when turning off the global middleware (You don't need to add this line if global middleware is enabled). Maybe a hint that this line is needed only when disabling global middleware would be best?

Works for me! Adding the value will not hurt in any case, but as you pointed out, could help reduce confusion. Thanks for making the adjustment ❤️

docs/content/3.application-side/5.guest-mode.md
```diff
@@ -15,6 +15,7 @@ This behavior is useful for login pages that you don't want to be visitable by l
 Briefly summarized, you can enable guest mode on a page by passing the following configuration:
 ```ts
 definePageMeta({
+  middleware: 'auth',
```
zoey-kaiser, line 18: Maybe we add a comment behind it indicating that this is not required if the `globalAppMiddleware` is enabled?

```suggestion
  middleware: 'auth', // Only required if globalAppMiddleware is disabled
```

zoey-kaiser (changes requested):

docs/content/3.application-side/5.guest-mode.md
anischihi: Thanks for the suggestion, took it in the commit 🙏

zoey-kaiser (approved):

PR #719, opened 2024-03-25
fix: opt in to `import.meta.*` properties

danielroe:
### ❓ Type of change

### 📚 Description

This is a very early PR to make this module compatible with [changes we expect to release in Nuxt v5](https://github.com/nuxt/nuxt/issues/25323).

In [Nuxt v3.7.0](https://github.com/nuxt/nuxt/releases/tag/v3.7.0) we added support for `import.meta.*` (see [original PR](https://github.com/nuxt/nuxt/pull/22428)) and we've been gradually updating docs and moving across from the old `process.*` patterned variables.

As I'm sure you're aware, these variables are replaced at build-time and enable tree-shaking in bundled code.
This change affects _runtime_ code (that is, that is processed by the Nuxt bundler, like vite or webpack) rather than code running in Node. So it really doesn't matter what the string is, but it makes more sense in an ESM-world to use `import.meta` rather than `process`.

(It might be worth updating the module compatibility as well to indicate it needs to have Nuxt v3.7.0+, but I'll leave that with you if you think this is a good approach.)

### 📝 Checklist


phoenix-ru:
@zoey-kaiser Let's re-schedule this to 0.8 to hint on possible breaking changes.

Would also be great to update our `Getting started -> Installation docs` to suggest a more concrete version than 
https://github.com/sidebase/nuxt-auth/blob/5d713aa419bcefc342f6a6e14899db6cc3d05d7f/docs/content/1.getting-started/2.installation.md?plain=1#L52
and also to make this more visible

phoenix-ru (approved):

zoey-kaiser (approved):

phoenix-ru (approved):

PR #740, opened 2024-04-18
chore: bump dependencies

phoenix-ru:
### ❓ Type of change

### 📚 Description
Security updates

### 📝 Checklist


docs/package.json
```diff
@@ -9,7 +9,7 @@
     "preview": "nuxi preview"
   },
   "devDependencies": {
-    "nuxt": "^3.0.0",
-    "@nuxt-themes/docus": "^1.0.0"
+    "@nuxt-themes/docus": "^1.15.0",
```
zoey-kaiser, line 12: We have to be careful with this update!

- As this version does not match the version of https://github.com/sidebase/docs, there may be features available here, that are not available in the real docs, which may lead to confusion
- They changed a lot with every docus version and they did not really follow semantic versioning, so updating may break a lot.

zoey-kaiser (approved):

docs/package.json
phoenix-ru: I explicitly checked the docus version before doing changes, it was `1.15.0`

Issue #411, opened 2023-05-26
Defining the redirect option for signin at the local provider causes an error.


akiot-b:
### Environment

"@sidebase/nuxt-auth": "^0.6.0-beta.2",
"nuxt": "^3.4.3",

I am in the process of migrating from @nuxtjs/auth-next and it is not working properly yet.
```
  auth: {
    provider: {
        type: "local",
        endpoints: {
          signIn: {
            path: '/api/auth/signin',
            method: 'post',
          },
          signOut: {
            path: '/api/auth/signout',
            method: 'post',
          },
          getSession: {
              path: '/api/auth/user',
            method: 'get',
          },
          // refresh: {
          //   url: '/api/auth/refresh',
          //   method: 'post',
          //   withCredentials: true,
          // },
        },
        pages: {
          login: '/'
        },
        token: {
          signInResponseTokenPointer: "/detail/access_token",
          type: "Bearer",
          headerName: "Authorization",
          maxAgeInSeconds: 60 * 15
        },
        sessionDataType: {
          email: 'string',
          name: 'string',
          password: 'string',
          user_roles: 'string[]'
        }
    },
    baseURL: 'http://127.0.0.1:3001/',
    // session: {
    //   enableRefreshPeriodically: true,
    //   enableRefreshOnWindowFocus: true,
    // },
    globalAppMiddleware: {
      isEnabled: false,
    },
  },
```

### Describe the bug

I am trying to SingIn with a local provider.
The default seems to be redirect: true, so I wanted to set redirect: false
````
await signIn({
        email: 'user@example.com',
        password: 'password', }
      }, { redirect:false })
````
If you look at the code, the redirect option (signInOptions) is passed as the body of the fetch destination. And in fact, it is objected as the body of the fetch along with email/password.
Is this correct behavior?

/@sidebase/nuxt-auth/dist/runtime/composables/local/useAuth.mjs

```
const signIn = async (credentials, signInOptions, signInParams) => {
  const nuxt = useNuxtApp();
  const config = useTypedBackendConfig(useRuntimeConfig(), "local");
  const { path, method } = config.endpoints.signIn;
  const response = await _fetch(nuxt, path, {
    method,
    body: {
      ...credentials,
      ...signInOptions ?? {}
    },
    params: signInParams ?? {}
  });
  const extractedToken = jsonPointerGet(response, config.token.signInResponseTokenPointer);
  if (typeof extractedToken !== "string") {
[...]

PR #755, opened 2024-05-21
fix: don't send sign-in options in request payload

despatates:
### 🔗 Linked issue

https://github.com/sidebase/nuxt-auth/issues/411

### ❓ Type of change

- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

When using method `signIn(credentials, { callbackUrl: '/' })` method with `local` or `refresh` provider, the `callbackUrl` and other options were sent to the server.  
As mentioned in https://github.com/sidebase/nuxt-auth/issues/411#issuecomment-1612376120, some servers can reject the request if the payload contains invalid parameters.

Those options are only used when the sign-in request is done, they shouldn't be sent to the server. They can be set directly in `credentials` data if they have to.

Flagged as _breaking change_ because some options may have to be copied from `signInOptions` to `credentials`.

### 📝 Checklist

### DIFF
--- a/src/runtime/composables/local/useAuth.ts
+++ b/src/runtime/composables/local/useAuth.ts
@@ -19,10 +19,7 @@ const signIn: SignInFunc<Credentials, any> = async (credentials, signInOptions, 
   const { path, method } = config.endpoints.signIn
   const response = await _fetch<Record<string, any>>(nuxt, path, {
     method,
-    body: {
-      ...credentials,
-      ...(signInOptions ?? {})
-    },
+    body: credentials,
     params: signInParams ?? {}
   })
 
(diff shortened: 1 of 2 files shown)


zoey-kaiser (approved):

Ulrich-Mbouna:
Yes i faced the same issue

sandrinejoy:
is this fixed?
I am still facing this issue when using signIn(payload, { external: true, redirect: false });
the options are send as the payload

despatates:
The issue is fixed but not yet released.
If you don't want to wait, install the `0.8.0-alpha.2` version (this is what I did).

PR #771, opened 2024-06-19
fix: Added getCurrentInstance check to conditionally register onMounted in useAuthState.

cip8:
Conditionally register onMounted within useAuthState to fix these warnings: 

```
onMounted is called when there is no active component instance to be associated with. Lifecycle injection APIs can only be used during execution of setup().
```


zoey-kaiser (dismissed):

zoey-kaiser:
Hi @cip8 👋 

Thanks for the PR! There is one extra semi-colan that fails in our listing. Could you please quickly adjust it so the linter can pass 😊

zoey-kaiser (approved):

Issue #551, opened 2023-10-18
When refresh browser on prerender env(SSG), Authentication status is lost.


artisthong:
### Environment

------------------------------
- Operating System: Darwin
- Node Version:     v18.17.1
- Nuxt Version:     3.7.4
- CLI Version:      3.9.0
- Nitro Version:    2.6.3
- Package Manager:  npm@9.6.7
- Builder:          -
- User Config:      devtools, build, alias, ssr, runtimeConfig, nitro, vite, modules, auth, i18n, experimental, css, pinia, app
- Runtime Modules:  @pinia/nuxt@0.4.11, @pinia-plugin-persistedstate/nuxt@1.1.2, @nuxtjs/i18n@8.0.0-rc.5, @sidebase/nuxt-auth@0.6.0-rc.0
- Build Modules:    -
------------------------------
client : npm run generate > npx serve [output static directory] -l 3001
node server :  npm run build(port 3000) > node index.mjs

### Reproduction

1. Call signIn on Login Page.
``` javascript
        const { signIn, status } = useAuth();
	setUser(<resLoginType>returnData.value.rs);
	// await authLogin();
	const wrapBody = {
		username: loginId.value,
		password: passwd.value,
		
	};
	await signIn(wrapBody, { redirect: false });

	navigateTo('/main', { replace: true });
```
2. signin is authorize and then move main page. (Screenshots : Chrome Inspector Application > Cookie and check useAuth status value) 
<img width="1069" alt="스크린샷 2023-10-18 오후 6 02 46" src="https://github.com/sidebase/nuxt-auth/assets/29880437/51c577df-62da-48ae-9f0e-7644ab553777">
<img width="187" alt="스크린샷 2023-10-18 오후 6 09 34" src="https://github.com/sidebase/nuxt-auth/assets/29880437/90dfbefe-4f6f-4051-b7cb-a6cfdab0798a">

3. After refreshing the browser, useAuth status value is lost and move the login.
<img width="320" alt="스크린샷 2023-10-18 오후 6 18 48" src="https://github.com/sidebase/nuxt-auth/assets/29880437/ed8290e2-d70f-431c-a16b-023f0c5519cb">

### Describe the bug

1. Authentication passed on the login page.
2. Refreshed the main page.
3. Authentication status is lost and moved to the login page.

I guess this is probably a problem that occurs when pre-rendering is performed on the client-driven server side.
Is there a solution to this?

### Additional context

 "nuxt.config.ts" part the below.
[...]

PR #785, opened 2024-07-03
updated fix authState getting wiped on page reload on static websites

Vijayabhaskar96:
set and reload data, rawToken from cookie

fix lint

fix typecheck

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/runtime/composables/local/useAuth.ts
+++ b/src/runtime/composables/local/useAuth.ts
@@ -9,7 +9,7 @@ import { formatToken } from '../../utils/local'
 import { useAuthState } from './useAuthState'
 // @ts-expect-error - #auth not defined
 import type { SessionData } from '#auth'
-import { useNuxtApp, useRuntimeConfig, nextTick, navigateTo } from '#imports'
+import { useCookie, useNuxtApp, useRuntimeConfig, nextTick, navigateTo } from '#imports'
 
 type Credentials = { username?: string, email?: string, password?: string } & Record<string, any>
 
@@ -91,12 +91,21 @@ const getSession: GetSessionFunc<SessionData | null | void> = async (getSessionO
   if (getSessionConfig) {
     const headers = new Headers(token ? { [config.token.headerName]: token } as HeadersInit : undefined)
     const { path, method } = getSessionConfig
+    const sessionCookie = useCookie<Object | null>('auth:sessionCookie', {
+      default: () => null,
+      maxAge: config.token.maxAgeInSeconds,
+      sameSite: config.token.sameSiteAttribute
+    })
 
     loading.value = true
     try {
       const result = await _fetch<any>(nuxt, path, { method, headers })
       const { dataResponsePointer: sessionDataResponsePointer } = config.session
       data.value = jsonPointerGet<SessionData>(result, sessionDataResponsePointer)
+      sessionCookie.value = {
+        lastRefreshedAt: lastRefreshedAt.value,
+        data: data.value
+      }
     } catch (err) {
       if (!data.value && err instanceof Error) {
         console.error(`Session: unable to extract session, ${err.message}`)
@@ -105,6 +114,7 @@ const getSession: GetSessionFunc<SessionData | null | void> = async (getSessionO
       // Clear all data: Request failed so we must not be authenticated
       data.value = null
       rawToken.value = null
+      sessionCookie.value = null
     }
     loading.value = false
     lastRefreshedAt.value = new Date()
--- a/src/runtime/plugin.ts
+++ b/src/runtime/plugin.ts
@@ -1,11 +1,13 @@
 import { getHeader } from 'h3'
 import authMiddleware from './middleware/auth'
 import { getNitroRouteRules } from './utils/kit'
-import { addRouteMiddleware, defineNuxtPlugin, useRuntimeConfig, useAuth, useAuthState, _refreshHandler } from '#imports'
+import { useTypedBackendConfig } from './helpers'
+import type { SessionCookie } from './types'
+import { addRouteMiddleware, defineNuxtPlugin, useRuntimeConfig, useAuth, useAuthState, useCookie, _refreshHandler } from '#imports'
 
 export default defineNuxtPlugin(async (nuxtApp) => {
   // 1. Initialize authentication state, potentially fetch current session
-  const { data, lastRefreshedAt, loading } = useAuthState()
+  const { data, lastRefreshedAt, rawToken, loading } = useAuthState()
   const { getSession } = useAuth()
 
   // use runtimeConfig
@@ -31,7 +33,27 @@ export default defineNuxtPlugin(async (nuxtApp) => {
 
   // Only fetch session if it was not yet initialized server-side
   if (typeof data.value === 'undefined' && !nitroPrerender && !disableServerSideAuth) {
-    await getSession()
+    // Restore the session data from the cookie if it exists
+    const config = useTypedBackendConfig(useRuntimeConfig(), 'local')
+    const sessionCookie = useCookie<SessionCookie | null>('auth:sessionCookie')
+    const cookieToken = useCookie<string | null>(config.token.cookieName)
+    if (sessionCookie.value && !rawToken.value && cookieToken.value) {
+      try {
+        loading.value = true
+        const sessionData = sessionCookie.value
+        lastRefreshedAt.value = sessionData?.lastRefreshedAt
+        data.value = sessionData?.data
+        rawToken.value = cookieToken.value
+      } catch (error) {
+        console.error('Failed to parse session data from cookie:', error)
+      } finally {
+        loading.value = false
+      }
+    }
+
+    if (!data.value) {
+      await getSession()
+    }
   }
 
   // 2. Setup session maintanence, e.g., auto refreshing or refreshing on foux
--- a/src/runtime/types.ts
+++ b/src/runtime/types.ts
@@ -573,3 +573,14 @@ export interface ModuleOptionsNormalized extends ModuleOptions {
     fullBaseUrl: string
   }
 }
+export interface SessionCookie {
+  lastRefreshedAt?: SessionLastRefreshedAt
+  data?: SessionDataObject
+}
+
+// Augment types
+declare module 'nuxt/schema' {
+  interface PublicRuntimeConfig {
+    auth: ModuleOptionsNormalized
+  }
+}


src/runtime/plugin.ts
```diff
@@ -31,7 +33,27 @@ export default defineNuxtPlugin(async (nuxtApp) => {
 
   // Only fetch session if it was not yet initialized server-side
   if (typeof data.value === 'undefined' && !nitroPrerender && !disableServerSideAuth) {
-    await getSession()
+    // Restore the session data from the cookie if it exists
```
zoey-kaiser, line 36: Changing this also results in changes to how the `authjs` provider works, which is causing the tests to fail. Please look into ensuring that your new logic does not effect this provider 😊

zoey-kaiser (changes requested):

src/runtime/plugin.ts
Vijayabhaskar96: fixed authjs failing problem. Sorry for ugly commit history, I'm not too familiar with git.

zoey-kaiser:
Hi @Geekimo 👋 

I would like to get a review by @phoenix-ru in before merging this PR 😊

Geekimo:
@zoey-kaiser Thanks for your reply ☺️

thorge:
Any news on this fix or can we help? I use a local provider and having auth work with ssg would be highly appreciated :-)

Geekimo:
Hello @zoey-kaiser and @phoenix-ru,
Sorry to bother you about this fix, but I'm currently holding a deployment to production due to this bug.
I'd really love to help you out to get this released.

Kind regards.

zoey-kaiser (dismissed):
Functionally approved 🥳

zoey-kaiser:
Hi @Vijayabhaskar96 👋 

I wanted to push this PR into the next release, but sadly the authjs tests are failing. We will finalize the PRs for the release tomorrow. For the case that we are not able to include it yet, you can directly install your PR as an npm package using:

```
pnpm add https://pkg.pr.new/@sidebase/nuxt-auth@785
```

src/runtime/plugin.ts
```diff
@@ -31,7 +33,29 @@ export default defineNuxtPlugin(async (nuxtApp) => {
 
   // Only fetch session if it was not yet initialized server-side
   if (typeof data.value === 'undefined' && !nitroPrerender && !disableServerSideAuth) {
-    await getSession()
+    // Restore the session data from the cookie
+    const config = useTypedBackendConfig(useRuntimeConfig(), 'local')
```
Vijayabhaskar96, line 37: authjs tests are failing because of some type mismatches in this line, I'm not much familiar with typescript, so can you take a look? @zoey-kaiser

Geekimo: @Vijayabhaskar96 Can you vive the error ? Maybe we can help.

Vijayabhaskar96: Thanks, but I have identified the problem and fixed it. Tests are passing.

Geekimo:
Hello @zoey-kaiser,
will this PR be included in the next release (if yes, when is it planned ?)
Regards

thorge:
Is there a specific reason why this wasn't included in the 0.9.4 release?

Geekimo:
Hello @zoey-kaiser,
The package that was built from this PR isn't working very well, there's import issues when try to call `useAuth()`.
Does someone have the same issue ?

```
[6:24:14 PM]  ERROR  Internal server error: Failed to resolve import "/Users/morgan/projects/***/node_modules/.pnpm/@sidebase+nuxt-auth@https+++pkg.pr.new+@sidebase+nuxt-auth@785_encoding@0.1.13_magicast@0.3.5_efzieaamfhvdbiybaqladlaygy/node_modules/@sidebase/nuxt-auth/dist/runtime/composables/refresh/useAuth" from "plugins/02.customFetch.ts". Does the file exist?
```

thorge:
Sorry, I don’t want to come across as impatient, but we’re currently holding back a production site until this fix is implemented. Is there already an idea of when this will be released?

src/runtime/types.ts
```diff
@@ -601,3 +601,14 @@ export interface ModuleOptionsNormalized extends ModuleOptions {
...
+  lastRefreshedAt?: SessionLastRefreshedAt
+  data?: SessionDataObject
+}
+
+// Augment types
+declare module 'nuxt/schema' {
+  interface PublicRuntimeConfig {
+    auth: ModuleOptionsNormalized
+  }
+}
```
phoenix-ru, line 614: Question as to why was this added?

src/runtime/composables/local/useAuth.ts
```diff
@@ -129,11 +129,22 @@ async function getSession(getSessionOptions?: GetSessionOptions): Promise<Sessio
...
   loading.value = true
   try {
     const result = await _fetch<any>(nuxt, path, { method, headers })
     const { dataResponsePointer: sessionDataResponsePointer } = config.session
     data.value = jsonPointerGet<SessionData>(result, sessionDataResponsePointer)
+
+    sessionCookie.value = {
+      lastRefreshedAt: lastRefreshedAt.value,
+      data: data.value
+    }
```
phoenix-ru, line 147: Storing user data in the cookie may lead to a GDPR violation if the `getSession` endpoint returns any personal information (which is often the case). It undoubtedly spawns the "protecting personal data" requirement for anyone using the module, even if they weren't affected by the SSG issue in the first place.

https://gdpr.eu/cookies/

I don't find it very attractive to introduce such a burden to NuxtAuth :/

src/runtime/plugin.ts
```diff
@@ -30,8 +32,52 @@ export default defineNuxtPlugin(async (nuxtApp) => {
...
+      lastRefreshedAt.value = sessionData?.lastRefreshedAt
+      data.value = sessionData?.data
+      rawToken.value = cookieToken.value
+    }
+    catch (error) {
+      console.error('Failed to parse session data from cookie:', error)
+    }
+    finally {
+      loading.value = false
+    }
```
phoenix-ru, line 80: I am struggling to understand as to why saving/restoring the session data in the cookie is needed in the first place? I read both the original issue #551 and the PRs around it, but still haven't found a valid reason for _this_ approach.

I think there should be a less hacky way of solving it

phoenix-ru:
@Geekimo @thorge @Vijayabhaskar96 Hi, sorry for keeping you waiting as this PR has been quite controversial for me. I just left my concerns in the review above. I don't want to give any false promises about merging the PR soon, as I do not feel confident at all in the patch.

If you trust it, you can use it at your risk by using the latest commit (see [the message from bot](https://github.com/sidebase/nuxt-auth/pull/785#issuecomment-2359018011) -> click on commit):
```
pnpm add https://pkg.pr.new/@sidebase/nuxt-auth@0f7e51a
```

I know it's one of the long-standing issues in the `local` provider and I will try to come back at it with a different approach within 2024.

src/runtime/composables/local/useAuth.ts
Geekimo: Maybe introducing an option to use SessionStorage nor LocalStorage would be great instead of using cookies ?

thorge: If an attacker gains access to a token stored in a Cookie, Session, or LocalStorage, the primary concern is not just identifying the user, but also the ability to access personal data and perform actions the user is authorized to do. This presents a clear security risk.

From a GDPR perspective, the critical question is whether the token can be used to track the user or if it contains personal data. Regardless of where the token is stored, if it can uniquely identify a user or track their actions, it falls under GDPR regulations.

For example, JWTs often store a user identifier in the payload. Since JWTs are not opaque and can be decoded, they may be used to directly identify a user, making them subject to GDPR. Even if the token does not explicitly store "personal data" like an email address, it qualifies as personal data if it can uniquely identify a user (e.g., through a user ID).

In this case, you must inform users via a cookie notice or a similar consent mechanism, ensuring transparency about the use of the token and obtaining their consent where required. However, if the cookies are strictly necessary for the operation of the web application (e.g., to keep users logged in or enable certain features), explicit consent may not be required under GDPR. This generally applies to authentication cookies, afaik.

src/runtime/plugin.ts
thorge: I encountered an issue with page reloads after logging in while using SSG with this module, causing the authentication token to be lost. It’s possible that I misconfigured something, but in general, I need a solution that ensures the user stays logged in even when he triggers a page reload.

When using SSG and wanting users to stay logged in across page reloads, client-side storage is required. I recommend using cookies for this purpose, as they are domain-specific, secure (with attributes like `HttpOnly` and `Secure`), and automatically sent to the server with each request, making them ideal for authentication.

localStorage is vulnerable to XSS. sessionStorage is cleared on page reloads, making it unsuitable for long-term session persistence.

While all three options have their trade-offs, cookies are typically the most secure and practical for authentication across page reloads imho. In case of JWT I guess the best option would be having short time tokens with long time opaque refresh tokens.

src/runtime/types.ts
Vijayabhaskar96: Sorry, I don't remember, but I guess it was causing merge conflicts.

src/runtime/plugin.ts
Vijayabhaskar96: One possible solution is to just load the raw token from the cookie and not store the session data anywhere.
Call `getSession` if `data.value` is null.

This would fetch the session every time the page is refreshed.

Will this work?

Vijayabhaskar96: The issue with the above approach is that middleware protects pages based on `status` - if a user 
refreshes a protected page, the `data.value` is lost, causing `status.value` to be 'unauthenticated'. 
This triggers unwanted redirects.

To prevent this, we can determine `status.value` based on the existence of `rawToken.value` for local 
provider only. This remains GDPR compliant as only the authentication token is stored in cookies, 
with the session being fetched on each page refresh.

Any other suggestions?

src/runtime/composables/local/useAuth.ts
phoenix-ru: @thorge You are mostly right. However, auth cookies/jwts don't need to store emails or other identifying information. A website can use "shallow" JWTs storing only user id in it (this is a valid `Access Token` in NuxtAuth terms). As to whether an ID is "personal data" is very subjective.

However if a website uses such a "shallow" JWT (`Access Token`) and makes a call to a `getSession` endpoint to check its validity + get user data, then it's a different thing. For example, internally we optimize such a call to also return user email, name, permissions, etc. on top of a regular JWT validity check. I would prefer to not store this information in a cookie.

Please note that I am also not clearly understanding the usecase of static websites fully. If there is authentication, what is the issue with the session being refetched on page reload?

src/runtime/plugin.ts
phoenix-ru: @Vijayabhaskar96 Thank you for a good explainer! I think this is the best take so far and I would absolutely be up to fixing unnecessary redirects (but without introducing any extra cookies - only using ones we have so far:

https://github.com/sidebase/nuxt-auth/blob/b1e4abad275c020daf8e227c1b4c9c7d1e485465/src/runtime/composables/local/useAuthState.ts#L36

https://github.com/sidebase/nuxt-auth/blob/b1e4abad275c020daf8e227c1b4c9c7d1e485465/src/runtime/composables/local/useAuthState.ts#L69

src/runtime/composables/local/useAuth.ts
thorge: @phoenix-ru There’s actually no need to store the session itself; we only need to store the access token. That’s what I thought this was about, but I must admit I didn’t examine the changes in detail. Sorry for my tldr answer above :-)

As @Geekimo and @Vijayabhaskar96 suggested, I’d prefer a solution where storing the access token in a cookie (or alternatively in SessionStorage/LocalStorage) is optional. If the session data isn’t available after a client page reload, the access token could be used to retrieve the session data as needed.

Issue #688, opened 2024-02-27
Is there a way to make authenticated requests out of the box?


PaulCombal:
### Describe the feature

I successfully managed to log in with Nuxt Auth! 

Now I want to query my API from the client with a REST request using the `Authentication: Bearer token` header.

I know Nuxt-Auth does this under the hood to get the context, right after logging in.

Does Nuxt Auth also provide a way to make our own authenticated requests? In the docs we have this example:

```
<script setup lang="ts">
const headers = useRequestHeaders(['cookie']) as HeadersInit
const { data: token } = await useFetch('/api/token', { headers })
</script>
```

Is this all? Is there a composable, or is this out of this library's scope? Is this the recommended way to make authenticated requests within the whole application?

Thank you

### Additional information

- [ ] Would you be willing to help implement this feature?

### Provider

- [ ] AuthJS
- [X] Local
- [ ] Refresh
- [ ] New Provider

DavidDeSloovere:
The docs say you have to pass the cookie explicitly for SSR, but it works without. I think Nuxt changed the default.


In the api, you can get the session or token with the composables, and you have enough to know who is making the call.
Unless your API is not in the nuxt application.

itsmohmans:
@DavidDeSloovere 
>  Unless your API is not in the nuxt application.

This is exactly the issue, I've used Nuxt Auth to sign in and get session data from an external API, but then I need to make requests like edit profile info and similar requests that need a valid token, so a composable provided by useAuth to make authenticated requests would be very helpful in such cases.

(I'm working on implementing this now but let me know if there's a better solution to this).

DavidDeSloovere:
I store the info/token from the external system in the nuxt auth token. Every call to the external API goes to Nuxt server routes (nitro). This is to keep the API key to the external service secret.

If you want to do that, here's something to get you started:


```typescript
// ./server/middleware/set-auth-context.ts
import type { Session } from "next-auth";
import type { JWT } from "next-auth/jwt";

import { getServerSession, getToken } from "#auth";

declare module "h3" {
  interface H3EventContext {
    authSession: Session | null;
    authToken: JWT | null;
    backendAccessToken: string | null;
  }
}
export default defineEventHandler(async (event) => {
  if (event.path.startsWith("/_")) {
    return;
  }

  const session = await getServerSession(event);
  event.context.authSession = session;

  const token = await getToken({ event: event });
  event.context.authToken = token;
  event.context.backendAccessToken
[...]

PR #787, opened 2024-07-04
Add a composable to make authenticated requests

itsmohmans:
### 🔗 Linked issue
#688

### ❓ Type of change

### 📚 Description
Add `authFetch` composable to provide a way to manually fetch a resource with authorization headers.

### 📝 Checklist

### DIFF
--- a/docs/content/3.application-side/2.session-access-and-management.md
+++ b/docs/content/3.application-side/2.session-access-and-management.md
@@ -62,6 +62,7 @@ const {
   signUp,
   signIn,
   signOut,
+  authFetch,
 } = useAuth()
 
 // Session status, either `unauthenticated`, `loading`, `authenticated`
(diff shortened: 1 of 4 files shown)


zoey-kaiser:
Hi @itsmohmans 👋 

Thank you for your contribution! While I think it is conducive to add these kinds of utilities to the module, I think we should approach it slightly differently:

Let us provide a more generic: `headers` composable that only returns the headers for the request. This would allow people to use the utility for other HTTP clients such as  [Axios](https://github.com/axios/axios). It would also avoid us having to hard-depend on the Nuxt `$fetch` methods and would also allow users to choose between using `$fetch,` `useFetch`, or `useAsyncData` without us writing custom wrappers for each.

Alternatively, I could also see use providing `header` and wrappers for all the fetch functionality inside Nuxt, however if we did this, I would propose adding a new composable to handle this. 

I will also pull in @phoenix-ru for his opinion on this matter!

itsmohmans:
Hey!
I agree it should be more generic; However, wouldn't a `headers` computed be somewhat similar to the current `token` property as we can just add it to the request headers along with anything else we want to add in the request?

I still do think the module needs a composable that makes authenticated requests out of the box, and I'm open for edits on the how we approach this in a better way.

PR #800, opened 2024-07-13
chore: use eslint flat config and add oxlint

zoey-kaiser:
### 🔗 Linked issue

closes N/A

### ❓ Type of change

- [X] 🧹 Chore (updates to the build process or auxiliary tools and libraries)

### 📚 Description

Updates our pretty outdated linting setup to make it more modern:

- Switch to Eslint flat config and use `@antfu/eslint-config` as a base
- Add OxLint to catch additional errors

### 📝 Checklist


zoey-kaiser:
@phoenix-ru I have noticed that the playground tests fail pretty quickly when updating the dependencies of the root `package.json`. This was also one of the reasons why I installed `vitepress` inside another `package.json` inside the `/docs` directory. 

In this PR I wanted to update our linter and installing oxlint already broke the tests. The error messages thrown by the tests are pretty crypic to me. Do you have any ideas on how we can make the tests rely less on the root `package.json` and be isolated to their own workspaces? I assume the issue comes from new dependency linkages that may use other versions of sub-dependencies then they do. However, as you are the experts on the tests, I wanted to pass this onto you, in case you has specific limitation / concerns about this 🤗

zoey-kaiser:
Waiting on #818

PR #820, opened 2024-07-19
docs: env value does not represent correct utilization of environment variables

NyllRE:
### 🔗 Linked issue
There isn't an issue but I didn't think I needed to open one since it's easy to resolve

### ❓ Type of change

- [X] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)

### 📚 Description
the nuxtAuthHandler page puts auth secret as hard coded string even though the documentation clearly wants the user to use the environment variables since it displays a tab with the `.env` file
```ts [nuxt.config.ts]
  runtimeConfig: {
    authSecret: '123',
  }
})
```
this is unsafe and might cause issues to unexperienced developers

### 📝 Checklist

- [X] I have updated the documentation accordingly.

### DIFF
--- a/docs/guide/authjs/nuxt-auth-handler.md
+++ b/docs/guide/authjs/nuxt-auth-handler.md
@@ -20,7 +20,7 @@ export default NuxtAuthHandler({
 ```ts [nuxt.config.ts]
 export default defineNuxtConfig({
   runtimeConfig: {
-    authSecret: '123',
+    authSecret: process.env.NUXT_AUTH_SECRET,
   }
 })
 ```


zoey-kaiser:
This is a false conception! By setting the environment variable `NUXT_AUTH_SECRET` you override `runtimeConfig.auth.secret`. We set a default `runtimeConfig.auth.secret`, however you can also set this to undefined. 

The issue with your change is that it would then require the `NUXT_AUTH_SECRET` to be set at build time and not runtime, which is not what most people want, as changing environment variables in the runtime is a lot easier to properly setup (and most environment variables are used at run time).

Please checkout the Nuxt docs on `useRuntimeConfig` for more information: https://nuxt.com/docs/guide/going-further/runtime-config

Issue #836, opened 2024-07-29
Unification and flexibility with `AUTH_ORIGIN`


hywax:
### Describe the feature

What problems there are now:
* Hardcode variable `AUTH_ORIGIN`. Not all projects are suitable for this name. 
  For example, if the project uses [NuxtSEO](https://nuxtseo.com/site-config/guides/setting-site-config), then `SITE_URL` should be specified there too, which in turn is equal to `AUTH_ORIGIN`. If `AUTH_ORIGIN` could be changed, then one variable would be used.
* It conflicts with Nuxt. Since Nuxt uses the `NUXT_` prefix (which can be changed).

### How would you implement this?

1) Switch completely to `useRuntimeConfig()`. This will get rid of naming inconsistencies
2) Add to `nuxt.config` the ability to set the variable dynamically

An example of how it will look like:
`nuxt.config.ts`
```ts
export default defineNuxtConfig({
  runtimeConfig: {
    authOrigin: '',
  },
  auth: {
    originEnvKey: 'AUTH_ORIGIN' // by default
  }
})
```
`.env`
```
NUXT_AUTH_ORIGIN=https://my.app
```

### Additional information

- [X] Would you be willing to help implement this feature?

### Provider

- [x] AuthJS
- [ ] Local
- [ ] Refresh
- [ ] New Provider

PR #818, opened 2024-07-18
feat(#673): migrate to AuthJS

phoenix-ru:
### 🔗 Linked issue

Closes #673 

### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [x] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have added tests (if possible).
- [ ] I have updated the documentation accordingly.

PR #837, opened 2024-07-29
feat(#836)!: unification and flexibility with `AUTH_ORIGIN`

hywax:
### 🔗 Linked issue

#836

### ❓ Type of change

### 📚 Description

Fully realized the whole idea of the issue.

### 📝 Checklist


pnpm-lock.yaml
zoey-kaiser, line 1: I see that you have upgraded the `pnpm` lockfile version to 9.x. Could you please undo this change and keep it on 6.x? We have already upgraded the version in another PR and noticed we needed to make a number of adjustments to the tests and CI for the upgrade to work properly 😊

zoey-kaiser (changes requested):
Hi @hywax 👋 

Thanks for the PR! I hope to get around to a full review this week. I already noticed one small issue I commented below!

pnpm-lock.yaml
hywax: I rebuilt the package. For such cases you should add `packageManager` field to `package.json`. Then pnpm will automatically install the dependencies of the required version.

phoenix-ru: `packageManager` field is actually on our todo list - will be added very shortly also to the `main` branch

src/runtime/server/services/utils.ts
```diff
@@ -9,14 +10,16 @@ import { useRuntimeConfig } from '#imports'
  * Get `origin` and fallback to `x-forwarded-host` or `host` headers if not in production.
  */
 export const getServerOrigin = (event?: H3Event): string => {
+  const authRuntimeConfig = useRuntimeConfig().public.auth
+
   // Prio 1: Environment variable
-  const envOrigin = process.env.AUTH_ORIGIN
+  const envOrigin = camelCase(authRuntimeConfig.originEnvKey!, { normalize: true })
```
phoenix-ru, line 16: I don't really get the point of this - you are camelCasing the env variable name? In the default case, it becomes
```ts
camelCase('AUTH_ORIGIN', { normalize: true }) // `authOrigin`
```

But this is not useful at all

src/runtime/types.ts
```diff
@@ -405,6 +405,15 @@ export interface ModuleOptions {
    * @default false
    */
   disableServerSideAuth?: boolean;
+  /**
+   * The environment key that holds the origin of the application. This is used to determine the full URL of the application in production.
```
phoenix-ru, line 409: The key doesn't actually hold anything - it only points to the env variable:
```suggestion
   * The name of the environment variable that holds the origin of the application. This is used to determine the full URL of the application in production.
```

phoenix-ru (changes requested):

docs/resources/error-reference.md
```diff
@@ -21,8 +21,18 @@ export default NuxtAuthHandler({
...
+```ts
+// file: nuxt.config.ts
+export default defineNuxtConfig({
+  runtimeConfig: {
+    authOrigin: 'https://example.org', // <--- !!!! THIS IS WHAT'S MISSING
+  }
+
+  // ... rest of your config
+})
+```
```
phoenix-ru, line 38: This is certainly not correct - `runtimeConfig.authOrigin` is never used

src/runtime/server/services/utils.ts
hywax: Nuxt waits for env variable with prefix `NUXT_` (by default, this prefix can be changed to a custom prefix). 

When Nuxt receives the variables, it converts them to camlCase by removing the `NUXT_` prefix. 

Here we get the ENV from the `authRuntimeConfig.originEnvKey` variable (whose name we can change to be package-independent)

`NUXT_AUTH_ORIGIN` -> `useRuntimeConfig().authOrigin`.

PS: i noticed, i'm not extracting the variable from ENV, i fixed that

Translated with DeepL.com (free version)

docs/resources/error-reference.md
hywax: Fixed it, it was a bug

src/runtime/server/services/utils.ts
hywax: Thanks for the comment, that was indeed a mistake

src/runtime/server/services/utils.ts
```diff
@@ -9,14 +10,17 @@ import { useRuntimeConfig } from '#imports'
  * Get `origin` and fallback to `x-forwarded-host` or `host` headers if not in production.
  */
 export const getServerOrigin = (event?: H3Event): string => {
+  const config = useRuntimeConfig()
+
   // Prio 1: Environment variable
-  const envOrigin = process.env.AUTH_ORIGIN
+  const envOriginKey = camelCase(config.public.auth.originEnvKey!, { normalize: true })
+  const envOrigin = config[envOriginKey] as string|undefined
```
phoenix-ru, line 17: It doesn't look correct to me - what would happen when a non-`NUXT_` env variable is used? I would assume that it is not picked up by Nuxt. After consulting with docs, it's exactly the case:

https://nuxt.com/docs/guide/going-further/runtime-config#environment-variables
> Only a specially-named environment variable can override a runtime config property. That is, an uppercase environment variable starting with NUXT_ which uses _ to separate keys and case changes.

I think that you still need to read from `process.env` regardless

```suggestion
  const envOriginKey = config.public.auth.originEnvKey!
  const envOriginKeyCamelcase = camelCase(envOriginKey, { normalize: true })
  const envOrigin: string | undefined = config[envOriginKeyCamelcase] ?? process.env[envOriginKey]
```

docs/guide/application-side/configuration.md
```diff
@@ -25,6 +26,13 @@ export default defineNuxtConfig({
 
 Whether the module is enabled at all
 
+## `originEnvKey`
+
+- **Type**: `string`
+- **Default**: `AUTH_ORIGIN`
+
+The environment variable that holds the origin of your application. This is used to determine the origin of your application in production. Read more [here](/resources/error-reference#auth-no-origin).
```
phoenix-ru, line 34:
```suggestion
The name of the environment variable that holds the origin of the application. This is used to determine the origin of your application in production. Read more [here](/resources/error-reference#auth-no-origin).
```

phoenix-ru (changes requested):

src/runtime/server/services/utils.ts
hywax: Yes, if there is no special prefix, the variable will be ignored. It should be so because we work in the Nuxt ecosystem. And it has a special mechanism of working with env variables. And bypassing it by communicating to process seems to me to be a mistake from the architectural point of view. 

Because it may cause side effects. Developers are used to the fact that an application uses variables with a prefix only. And they may not realize that such a variable already exists in the system, it will cause side effects and difficulties.

I am very much against this approach, but if you are ready that side effects appear in the application because of accessing process, then fine.

hywax: @phoenix-ru leave things as they are? Or accept your changes?

hywax:
@zoey-kaiser break-change tag can be removed as backward compatibility is preserved

src/runtime/server/services/utils.ts
phoenix-ru: Issue with only accepting `NUXT_` prefixes is that some providers (like Vercel/Netlify) set their env variables automatically for preview envs. And it has been quite a requested feature to support them.

By using `process.env` we allow devs to specify `originEnvKey: 'VERCEL_URL'` and NuxtAuth will use this variable.

> Because it may cause side effects

I completely agree with you here and would also love to only use `NUXT_` prefixed env variables and completely ditch `process.env`, but in this case I think we should accommodate for different use-cases. This is also done because `next-auth` itself relies on `process.env`:

https://github.com/nextauthjs/next-auth/blob/064ee18164bbb479bf9727361088f4a8b3738526/packages/core/src/lib/utils/env.ts#L27-L33

phoenix-ru (dismissed):

phoenix-ru:
@hywax thank you for taking your time to push this PR to the end, it's very appreciated.
I think in general it is in the right direction and I feel confident enough with merging it and picking up on your work to bring some improvements to origin detection discovered in #818

docs/guide/advanced/deployment/self-hosted.md
```diff
@@ -6,7 +6,7 @@ This guide will explain how you can self-host a Nuxt3 application running NuxtAu
 
 When deploying the Auth.JS provider, the application must be informed what URL it is running at. This is to properly determine callback urls when navigating users to external OAuth providers. Depending on your setup, NuxtAuth allows you to set this value at either [**Runtime**](https://nuxt.com/docs/guide/going-further/hooks#app-hooks-runtime) or [**Buildtime**](https://nuxt.com/docs/guide/going-further/hooks#nuxt-hooks-build-time).
 
-- **Runtime:** Set the `AUTH_ORIGIN` environment variable.
+- **Runtime:** Set the `NUXT_AUTH_ORIGIN` environment variable.
```
zoey-kaiser, line 9: Is the default now set to: `NUXT_AUTH_ORIGIN` or just `AUTH_ORIGIN`? If it gets changed to `NUXT_AUTH_ORIGIN`, I belive it is still a breaking change. Otherwise are the docs just outdated here?

docs/resources/error-reference.md
```diff
@@ -21,8 +21,18 @@ export default NuxtAuthHandler({
...
 2. Development only: Determine the origin automatically from the incoming HTTP request
 
-
 The `origin` is important for callbacks that happen to a specific origin for `oauth` flows. Note that in order for (2) to work the `origin` already has to be set at build-time, i.e., when you run `npm run build` or `npm run generate` and it will lead to the `origin` being inside your app-bundle.
+
+```ts
+// file: nuxt.config.ts
+export default defineNuxtConfig({
+  runtimeConfig: {
+    authOrigin: 'https://example.org', // <--- !!!! THIS IS WHAT'S MISSING
```
zoey-kaiser, line 33:
```suggestion
    authOrigin: 'https://example.org', // You can either set a default or leave it empty
```

zoey-kaiser (changes requested):

docs/guide/advanced/deployment/self-hosted.md
hywax: `AUTH_ORIGIN` support remains, but `NUXT_AUTH_ORIGIN` is preferred. Therefore, it is suggested to use `NUXT_AUTH_ORIGIN` for new users

phoenix-ru: The default looked to me like `AUTH_ORIGIN` like we have now, isn't it?

zoey-kaiser: > AUTH_ORIGIN support remains, but NUXT_AUTH_ORIGIN is preferred. Therefore, it is suggested to use NUXT_AUTH_ORIGIN for new users

Okay! So by default both are supported?

hywax: Yes :)

There is this line in the config:
```
originEnvKey: 'AUTH_ORIGIN',
```
This means that the ENV will be `NUXT_AUTH_ORIGIN` or `AUTH_ORIGIN` by default. The user will be able to change it if they want to

phoenix-ru (approved):
I created a follow-up issue. The PR can be merged in its current state

Issue #848, opened 2024-08-08
Polish `NUXT_AUTH_ORIGIN` logic


phoenix-ru:
Follow up on #837

              Does not need to be done in this PR. but as a follow-up let's:

- Adapt the `~/playground-*` to use `NUXT_AUTH_ORIGIN`
- Adapt any starter to use `NUXT_AUTH_ORIGIN`
- Adapt the nuxt-auth-example to use `NUXT_AUTH_ORIGIN`

_Originally posted by @zoey-kaiser in https://github.com/sidebase/nuxt-auth/pull/837#pullrequestreview-2227333420_
            

Additionally:
- [x] Change default `originEnvKey` to `NUXT_AUTH_ORIGIN` and automatically strip `NUXT_` prefix for camel-case config keys
- [x] Check type of variable from `config` to be `string | undefined` - report error otherwise

Issue #673, opened 2024-02-23
Migration to authjs 🚀 


zoey-kaiser:
Hello everyone 👋 

The time has come: We are slowly moving forward to migrate NuxtAuth from using NextAuth under the hood to running the new authjs package!

This will come with numerous benefits, including:
- Smaller bundle size (as we no longer need to include the default Login Page built for React)
- Improvements to many bugs that have been blocked by NextAuth
- Support for future features
	- Serverless deployments
	- etc.

This issue will track our current progress, issues and goals. We are currently beginning this migration, therefore more information will follow soon.

phoenix-ru:
Valid point regarding `trustHost`: https://github.com/sidebase/nuxt-auth/issues/691#issuecomment-1972763234

phoenix-ru:
It makes perfect sense to wait for a release of `next-auth@5` which would also mean a release of a significant `@auth/core` version (maybe even v1).

Tracking:
https://github.com/nextauthjs/next-auth/releases

zoey-kaiser:
> It makes perfect sense to wait for a release of next-auth@5 which would also mean a release of a significant @auth/core version (maybe even v1).

Agreed! I can see that they already have alpha versions of Version 5 released. Have they added support for `trusthost` there yet? I think we can also still continue internally preparing for the release, by outlining which changes we will be making and updating our internal logic in preparation for this. But I do agree that we should wait with the official switch until they release!

hopkins385:
I am a bit concerned, see reports below.
What is missing/required to start with the migration process?

package.json
```
"@sidebase/nuxt-auth": "^0.7.2",
"next-auth": "4.21.1",
```

Audit Report:
```shell
# npm audit report
next  >=13.4.0 <14.1.1
Severity: high
Next.js Server-Side Request Forgery in Server Actions - https://github.com/advisories/GHSA-fr5h-rqp8-mj6g
fix available via `npm audit fix`
node_modules/next

next-auth  <4.24.5
Severity: moderate
Possible user mocking that bypasses basic authentication - https://github.com/advisories/GHSA-v64w-49xw-qq89
fix available via `npm audit fix --force`
Will install @sidebase/nuxt-auth@0.6.7, which is a breaking change
node_modules/next-auth
  @sidebase/nuxt-auth  0.3.0-alpha.1 - 0.4.0-alpha.6 || >=0.7.0-rc.0
  Depends on vulnerable versions of next-auth
[...]

Issue #523, opened 2023-08-31
getServerSession mutates event argument


j2L4e:
### Environment

------------------------------
- Operating System: Linux
- Node Version:     v18.16.0
- Nuxt Version:     3.7.0
- CLI Version:      3.7.3
- Nitro Version:    2.6.2
- Package Manager:  npm@9.5.1
- Builder:          -
- User Config:      css, ssr, devtools, modules, auth, vuetify, runtimeConfig
- Runtime Modules:  @invictus.codes/nuxt-vuetify@0.3.0, @sidebase/nuxt-auth@0.6.0-beta.5, @pinia/nuxt@0.4.11
- Build Modules:    -
------------------------------

### Reproduction

Call getServerSession() multiple times from within any api route handler

```js
// /server/api/bug.ts

import { getServerSession } from "#auth";

export default defineEventHandler(async (event) => {
    console.log(event.node.res.getHeaders())
    // [Object: null prototype] {}

    await getServerSession(event)
    console.log(event.node.res.getHeaders())
    // [Object: null prototype] { 'content-type': 'application/json' }

    await getServerSession(event)
    console.log(event.node.res.getHeaders())
    // [Object: null prototype] {
    //   'content-type': [ 'application/json', 'application/json' ]
    // }

    return []
})
```

### Describe the bug

Calling `getServerSession` internally calls the auth handler. This has side-effects because cookies and headers are set. 
In my case, having multiple `Content-Type` headers set, caused errors in client-side content type detection. `$fetch` wouldn't automatically parse `application/json` responses correctly, but return a blob instead.

Some component of the nuxt dev setup seems to be deduping or resetting `Content-Type` headers, effectively mitigating this problem in development. an observable error only occurs in production builds.

The cause of this can be observed in dev mode, though.

xak2000:
I have a similar problem caused by this behavior.

I have a Nitro middleware that calls `getServerSession(event)`. If an actual API handler also calls `getServerSession(event)`, then `useFetch()` doesn't read data properly when it is called from SSR. It decodes the data as `Blob` instead of simple JS `object`, so the client can't render the data.

So, calling `getServerSession(event)` more than one time from anywhere (e.g. a handler, a middleware) causes this problem.

The problem is caused by modifying the response headers. Even calling `getServerSession(event)` just once still modifies the headers, so this can still cause the same problem if the response headers already have had a `content-type` header at the moment of the first call of `getServerSession(event)`.

### Examples:

[...]

PR #849, opened 2024-08-08
feat(#673, #523, #848): back-port authjs migration

phoenix-ru:
### 🔗 Linked issue

Relates to #673, #523, #848

This is a (partial) back-port of #818 to current code. Because completely updating to `@auth/core` makes the module very unstable, we decided to instead back-port the advancements made in the PR to also prepare for a future update.

Differences with #818 :
- no `@auth/core` version bump;
- internal helpers are preserved because they seem to work better than these of `@auth/core`;
- no breaking changes.

### ❓ Type of change

### 📝 Checklist


zoey-kaiser (changes requested):
- [ ] When I started the playground I got the following warning spammed in the console (c1). Note: This only occurs when a unauthenticated user visits the site


C1:
```
 WARN  [nuxt] useAsyncData must return a value (it should not be undefined) or the request may be duplicated on the client side.
```

phoenix-ru:
> When I started the playground I got the following warning spammed 

This is likely related to dependencies. I will run an additional update cycle and see if that resolves the problem for you

phoenix-ru:
Documented the issue here (tldr; it is not caused by us):
https://github.com/nuxt/nuxt/issues/28096#issuecomment-2291100566

zoey-kaiser (approved):
Thanks for the investigation! The workaround does remove the warning message. I see no reason to move forward with merging this PR. 

However, let's discuss how this PR integrates into our planned migration to authjs and what the next steps are prior to merging 😊

PR #853, opened 2024-08-09
chore: updated ESLint and general housekeeping

zoey-kaiser:
### 🔗 Linked issue

Closes N/A

### ❓ Type of change

### 📚 Description

This PR:
- Removed old non-flat eslint setup
- Installed Antfu Eslint Config and sets up our usual development rules
- Adds `.vscode` folder for extensions and settings
- Installs OXLint
- Fixes all issues that the new linters discovered

### 📝 Checklist


src/runtime/composables/authjs/useAuth.ts
```diff
@@ -138,6 +140,7 @@ const signIn: SignInFunc<SupportedProviders, SignInResult> = async (provider, op
 
   // At this point the request succeeded (i.e., it went through)
   const error = new URL(data.url).searchParams.get('error')
+  // eslint-disable-next-line ts/no-use-before-define
```
zoey-kaiser, line 143: Note: For some issues I decided to disable the eslint rule for now. The biggest Issue that I did this for was when functions were used before they were defined.

This change would require making them all top-level functions (which I think is a good idea) but sadly we have some helpers (e.g. `makeCWN`: Make a call with Nuxt function) that makes this a bit more tricky. In addition to this, we also type a lot of functions with their params and returns, which was also tricky to continue using this, while making them top-level functions. 

@phoenix-ru, feel free to fix these issues if you would like in this PR or a follow-up PR. However, I think its good to avoid big code changes here and only address minimal changes to avoid any breaking changes!

zoey-kaiser:
We also disabled all Nuxt Auto-imports. This was done a long time ago, I was wondering if we could maybe re-enable them, as this could reduce the amount of imports (+ incorrect typings) and also make the playgrounds act more like an actual Nuxt App would!

https://github.com/sidebase/nuxt-auth/blob/main/.nuxtrc#L1

phoenix-ru (approved):

Issue #834, opened 2024-07-26
enableOnWindowFocus must be used only in authenticated condition


YoshimiShima:
### Environment

- Operating System: Darwin
- Node Version:     v20.11.1
- Nuxt Version:     3.12.2
- CLI Version:      3.12.0
- Nitro Version:    2.9.6
- Package Manager:  pnpm@9.5.0
- Builder:          -
- User Config:      nitro, app, components, debug, devServer, devtools, dir, experimental, googleFonts, i18n, ignore, imports, modules, routeRules, build, runtimeConfig, sourcemap, srcDir, tailwindcss, testUtils, typescript, hooks, vite, auth
- Runtime Modules:  @pinia/nuxt@0.5.1, @vueuse/nuxt@10.11.0, @nuxt/test-utils/module@3.13.1, @nuxtjs/google-fonts@3.2.0, @nuxtjs/i18n@8.3.1, @nuxtjs/tailwindcss@6.12.0, @vee-validate/nuxt@4.13.1, @sidebase/nuxt-auth@0.8.1, @pinia-plugin-persistedstate/nuxt@1.2.1, @nuxt/eslint@0.3.13
- Build Modules:    -

### Reproduction

```ts
// nuxt.config.ts

provider: {
  type: 'refresh',
  pages: {
    login: '/login',
  },
},
 sessionRefresh: {
   enablePeriodically: 90000,
   enableOnWindowFocus: true,
 },
globalAppMiddleware: {
   isEnabled: true,
},
```

```ts
// login.vue

definePageMeta({
   auth: {
     unauthenticatedOnly: true,
   },
})
```

### Describe the bug

The enableOnWindowFocus feature should not be triggered on pages that are accessible to unauthenticated users. Currently, when a user switches back to a tab containing an unauthenticated page (such as the login page of us), the refresh mechanism is still executed, which is unnecessary and potentially problematic.

In the current implementation in `src/runtime/utils/refreshHandler.ts`, there's an inconsistency in how `enablePeriodically` and `enableOnWindowFocus` are handled.
For `enablePeriodically`, there's a check for the authenticated state:

https://github.com/sidebase/nuxt-auth/blob/9295d1fe4f27d42c879862a48202f5b215fc0048/src/runtime/utils/refreshHandler.ts#L35-L42

However, for `enableOnWindowFocus`, no such check exists:
[...]

PR #858, opened 2024-08-11
fix(#834): Do not refresh on window focus for unprotected pages

YoshimiShima:
### 🔗 Linked issue

#834 

### ❓ Type of change

### 📚 Description
The enableOnWindowFocus feature should not be triggered on pages that are accessible to unauthenticated users. Currently, when a user switches back to a tab containing an unauthenticated page (such as the login page of us), the refresh mechanism is still executed, which is unnecessary and potentially problematic.
This pull request addresses that issue by ensuring that the enableOnWindowFocus feature is only executed when the user is in an authenticated state. This change prevents unnecessary refreshes on unauthenticated pages and eliminates potential problems associated with triggering the refresh mechanism on pages accessible to unauthenticated users.

### 📝 Checklist

### DIFF
--- a/src/runtime/utils/refreshHandler.ts
+++ b/src/runtime/utils/refreshHandler.ts
@@ -73,7 +73,7 @@ export class DefaultRefreshHandler implements RefreshHandler {
     // Listen for when the page is visible, if the user switches tabs
     // and makes our tab visible again, re-fetch the session, but only if
     // this feature is not disabled.
-    if (this.config?.enableOnWindowFocus && document.visibilityState === 'visible') {
+    if (this.config?.enableOnWindowFocus && document.visibilityState === 'visible' && this.auth?.status.value !== 'unauthenticated') {
       this.auth?.refresh()
     }
   }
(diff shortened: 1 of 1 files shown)


src/runtime/utils/refreshHandler.ts
```diff
@@ -73,7 +73,7 @@ export class DefaultRefreshHandler implements RefreshHandler {
     // Listen for when the page is visible, if the user switches tabs
     // and makes our tab visible again, re-fetch the session, but only if
     // this feature is not disabled.
-    if (this.config?.enableOnWindowFocus && document.visibilityState === 'visible') {
+    if (this.config?.enableOnWindowFocus && document.visibilityState === 'visible' && this.auth?.status.value !== 'unauthenticated') {
```
zoey-kaiser, line 76: Above, the check to refresh periodically, is based on if `this.auth?.data.value` is set. Do you see any advantage in checking against the status? 

I think we should use the same check for both, to ensure consistency, so I would propose to either change this to check if the data is set, or change the check for `refreshPeriodically`, if you belive checking against status is better 😊

zoey-kaiser (changes requested):

src/runtime/utils/refreshHandler.ts
YoshimiShima: @zoey-kaiser 
Hi zoey🫶
Thank you for your kind review! Yes you are right. I've confirmed how it works with using `data.value` in 4fd1da8f1452ab0e174b461ccd59edf32111a571 
please kindly confirm😊

zoey-kaiser (approved):

Issue #861, opened 2024-08-13
Redirect back to requested page after authentication


andychukse:
### Describe the feature

Redirect user back to requested page after authentication
Add redirect query parameter to middleware navigateTo when a guest visits authenticated page. 
This will help users redirect back to previous page after sign in.

```
// nuxt.config.ts

provider: {
  type: 'local',
  refresh: { 
        isEnabled: true, 
        endpoint: { path: '/refresh', method: 'POST' }, 
        refreshOnlyToken: true
   },
  pages: {
    login: '/login',
  },
},
 sessionRefresh: {
   enablePeriodically: 90000,
   enableOnWindowFocus: true,
 },
globalAppMiddleware: {
   isEnabled: true,
   addDefaultCallbackUrl: '/account'
},
```

```
// login.vue

definePageMeta({
   auth: {
     unauthenticatedOnly: true,
   },
})
```
Currently when guest tries to access a protected page they are redirected to `/login` page and after login they are redirect to default callbackUrl unless they specify callbackUrl in the `signIn` function.

This feature will help automatically redirect a user back to the page they requested after login. For example, if they requested to access `'/profile'` and they are redirected to login page, after successful login, they will be redirected back to `'/profile'`

### How would you implement this?

Current code
Current implementation redirects to login page

https://github.com/sidebase/nuxt-auth/blob/de1dca620a88b1cec2b408bbe27b67aacd72c355/src/runtime/middleware/auth.ts#L100-L103

Modified code
Proposed implementation gets the current url page and append as redirect query to the login page

```
return navigateTo({
      path: authConfig.provider.pages.login,
      query: {
        redirect: to.fullPath,
      },
    })
  }
```

https://github.com/andychukse/nuxt-auth/blob/f3ccf913a9121ac874b30c8b4a9a6e534a036d36/src/runtime/middleware/auth.ts#L108C5-L113C7

Current signIn function or local and refresh providers checks if callbackUrl is set, if not get the default callbackUrl

https://github.com/sidebase/nuxt-auth/blob/de1dca620a88b1cec2b408bbe27b67aacd72c355/src/runtime/composables/local/useAuth.ts#L41-L47

To handle the redirect on refresh Provider and local provider.
Check for the presence of redirect query parameter on the url and navigate to that if no callbackUrl is set. If no redirect query parameter is detected, then fallback to the default callbackUrl.

implementation in local and refresh provider

```
if (typeof callbackUrl === 'undefined') {
[...]

PR #870, opened 2024-08-19
enh(#861): Redirect back to requested page after authentication

andychukse:
### 🔗 Linked issue
#861 

### ❓ Type of change

### 📚 Description
Add functionality redirect user back to requested page after authentication
Added a redirect query parameter to auth middleware navigateTo when a guest visits authenicated page.
This will help users redirect back to previous page after sign in.
Currently when guest tries to access a protected page they are redirected to `/login` page and after login they are redirect to default `callbackUrl `unless they specify `callbackUrl` in the `signIn` function.

This feature will help automatically redirect a user back to the page they requested after login. For example, if they requested to access `'/profile'` and they are redirected to login page, after successful login, they will be redirected back to `'/profile'`. They can also override this by providing a `callbackUrl `in the signIn function as it is currently

### 📝 Checklist

### DIFF
--- a/src/runtime/composables/local/useAuth.ts
+++ b/src/runtime/composables/local/useAuth.ts
@@ -10,7 +10,7 @@ import { type UseAuthStateReturn, useAuthState } from './useAuthState'
 import { callWithNuxt } from '#app/nuxt'
 // @ts-expect-error - #auth not defined
 import type { SessionData } from '#auth'
-import { navigateTo, nextTick, useNuxtApp, useRuntimeConfig } from '#imports'
+import { navigateTo, nextTick, useNuxtApp, useRoute, useRuntimeConfig } from '#imports'
 
 type Credentials = { username?: string, email?: string, password?: string } & Record<string, any>
 
@@ -60,7 +60,12 @@ const signIn: SignInFunc<Credentials, any> = async (credentials, signInOptions, 
   const { redirect = true, external } = signInOptions ?? {}
   let { callbackUrl } = signInOptions ?? {}
   if (typeof callbackUrl === 'undefined') {
-    callbackUrl = await determineCallbackUrl(runtimeConfig.public.auth, () => getRequestURLWN(nuxt))
+    if (useRoute()?.query?.redirect) {
+      callbackUrl = useRoute().query.redirect?.toString()
+    }
+    else {
+      callbackUrl = await determineCallbackUrl(runtimeConfig.public.auth, () => getRequestURLWN(nuxt))
+    }
   }
   if (redirect) {
     return navigateTo(callbackUrl, { external })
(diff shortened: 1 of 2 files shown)


YarikSWT:
Really looking forward to this feature

zoey-kaiser (changes requested):
Love this feature! Thank you so much for your PR ❤️ 

One suggestion I would have is to respect the [`addDefaultCallbackUrl`](https://github.com/sidebase/nuxt-auth/blob/main/src/runtime/types.ts#L36) option inside the middleware and only add the callback URL to the redirect if it is enabled! 

This should be a pretty straight forward adjustments, where you check if [`globalAppMiddleware.addDefaultCallbackUrl`](https://github.com/sidebase/nuxt-auth/blob/main/src/runtime/middleware/auth.ts#L89) is set before adding `query.redirect` to the `navigateTo` call!

src/runtime/middleware/auth.ts
```diff
@@ -105,6 +105,11 @@ export default defineNuxtRouteMiddleware((to) => {
     return navigateTo(metaAuth.navigateUnauthenticatedTo)
   }
   else {
-    return navigateTo(authConfig.provider.pages.login)
+    return navigateTo({
+      path: authConfig.provider.pages.login,
+      query: {
+        redirect: to.fullPath
```
zoey-kaiser, line 111: Only add this if `globalAppMiddleware.addDefaultCallbackUrl` is `true` (Note: it can also be a string, if this is the case, add that string as the redirect` path)

andychukse: Thanks @zoey-kaiser ! I have updated as you suggested.

zoey-kaiser (dismissed):
Functionally approved 🥳

@phoenix-ru, please do a code review 😊

src/runtime/composables/local/useAuth.ts
```diff
@@ -60,7 +60,12 @@ const signIn: SignInFunc<Credentials, any> = async (credentials, signInOptions,
   const { redirect = true, external } = signInOptions ?? {}
   let { callbackUrl } = signInOptions ?? {}
   if (typeof callbackUrl === 'undefined') {
-    callbackUrl = await determineCallbackUrl(runtimeConfig.public.auth, () => getRequestURLWN(nuxt))
+    if (useRoute()?.query?.redirect) {
+      callbackUrl = useRoute().query.redirect?.toString()
```
phoenix-ru, line 64:
```suggestion
      callbackUrl = useRoute().query.redirect.toString()
```

phoenix-ru (changes requested):

phoenix-ru (approved):
@andychukse Thanks for your PR! I think it will bring `local` provider in parity with what `authjs` already does in terms of redirects

zoey-kaiser:
PR Released in https://github.com/sidebase/nuxt-auth/releases/tag/0.9.3 🥳

Issue #890, opened 2024-08-29
Infinite loop on refresh token endpoint


Suniron:
### Environment

------------------------------
- Operating System: Linux
- Node Version:     v20.17.0
- Nuxt Version:     3.13.0
- CLI Version:      3.13.1
- Nitro Version:    2.9.7
- Package Manager:  pnpm@9.0.4
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

With the configuration below, sign in to the app and reload the page

```ts
const determineBaseUrl = () => {
  const baseUrl = process.env.NUXT_PUBLIC_BASE_API || '/api'
  // Check if the baseUrl finishes with a trailing slash
  return !baseUrl.endsWith('/') ? `${baseUrl}/auth/` : `${baseUrl}auth/`
}

const authOptions: ModuleOptions = {
  baseURL: determineBaseUrl(),
  globalAppMiddleware: true,
  isEnabled: true,
  provider: {
    endpoints: {
      getSession: { method: 'get', path: `session` },
      signIn: { method: 'post', path: `credentials` },
      signOut: { method: 'delete', path: `logout` },
      // @ts-expect-error disable sign up, like in the example from the docs: https://auth.sidebase.io/guide/local/quick-start#api-endpoints
      signUp: false,
    },
    pages: {
      login: '/login',
    },
    // See: https://github.com/sidebase/nuxt-auth/issues/867#issuecomment-2293906780
    refresh: {
      endpoint: { method: 'post', path: `refresh-token` },
      isEnabled: true,
      refreshOnlyToken: true,
      token: {
        cookieName: 'auth.refresh',
        httpOnlyCookieAttribute: false,
        maxAgeInSeconds: 2592000,
        refreshRequestTokenPointer: '/refreshToken',
        sameSiteAttribute: 'lax',
        secureCookieAttribute: false,
        signInResponseRefreshTokenPointer: '/refreshToken',
      },
    },
    session: {
      dataType: {
        companyHasAcceptedTermsOfUse: 'boolean',
        companyId: 'number',
        companyName: 'string',
        email: 'string',
        firstName: 'string',
        fullyConnected: 'boolean',
        id: 'string',
        isTwoFactorInitialized: 'boolean',
        lastName: 'string',
        roles: '(\'admin\' | \'member\')[]',
        username: 'string',
      },
    },
    token: {
      cookieName: 'auth.token',
      headerName: 'Authorization',
[...]

PR #902, opened 2024-09-06
fix: Refresh not triggered if the access token is not also known, which limits its purpose

cip8:
### 🔗 Linked issue

#890 [second part]

### ❓ Type of change

### 📚 Description

The refresh logic should be triggered if the refresh token is known.

Some backends might require both tokens for a refresh, but this is not default behavior.

Refresh tokens are long-lasting and usually expire after access tokens, so requiring both for a successful refresh should be optional. 

### 📝 Checklist

### DIFF
--- a/src/runtime/composables/local/useAuth.ts
+++ b/src/runtime/composables/local/useAuth.ts
@@ -10,7 +10,7 @@ import { type UseAuthStateReturn, useAuthState } from './useAuthState'
 import { callWithNuxt } from '#app/nuxt'
 // @ts-expect-error - #auth not defined
 import type { SessionData } from '#auth'
-import { navigateTo, nextTick, useNuxtApp, useRuntimeConfig } from '#imports'
+import { navigateTo, nextTick, useCookie, useNuxtApp, useRuntimeConfig } from '#imports'
 
 type Credentials = { username?: string, email?: string, password?: string } & Record<string, any>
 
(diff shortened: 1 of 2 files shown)


cip8:
Fix is incomplete - logout doesn't clear refresh cookie so the user re-authenticates on each visit.
Will update soon.

cip8:
Tested with my local setup, everything works fine now.

src/runtime/composables/local/useAuth.ts
```diff
@@ -87,6 +87,28 @@ const signOut: SignOutFunc = async (signOutOptions) => {
...
 
+  // Explicitly clear both cookies to avoid watcher issues
+  // Use the same options to clear the cookie as the ones used to set it
+  useCookie(config.token.cookieName, {
+    maxAge: 0,
+    domain: config.token.cookieDomain,
+    sameSite: config.token.sameSiteAttribute,
+    secure: config.token.secureCookieAttribute,
+    httpOnly: config.token.httpOnlyCookieAttribute
+  }).value = null
```
phoenix-ru, line 98: I've been trying to understand why is this required - could you please clarify that?

einz-loggik:
@cip8 Whats your status on this? Would love to see this merged because it should fix the problem with automatic logout if the access token reached EOL.

cip8:
Hey @einz-loggik 👋 

Yeah, I’d really like to see this merged as well - but I’ve decided to move on from contributing to this project.

I’ve created my own fork and use it for my projects instead. 

The incredibly long wait times for reviews & some challenges with its maintainers just don't align with how I like to work.

Good luck 💙🖖

src/runtime/composables/local/useAuth.ts
cip8: coding is hard! for skill issues, use chatgpt

Issue #843, opened 2024-08-02
return data from Signup() Local Composable


iamKiNG-Fr:
### The  Feature
To allow `SignUp()` useAuth local composable return data from signup request.

### The Journey
I'm unable to get the return data from the Signup request I send. This request contains data that is important for my applications signup process.

This puzzled me for a while after I tried assigning a variable to the `signup()` sidebase useAuth() function and then log that variable, only to  get an `undefined` value even when i could see the response with the dev tools.

```
/pages/register.vue
const response = await signUp(credentials, undefined, { preventLoginFlow: true })
console.log(response);
// undefined
```

**chrome dev tool**

![image](https://github.com/user-attachments/assets/191d5e23-6eba-497c-a422-263745fda786)

### The Discovery

After further investigation, i noticed that the sidebase nuxt-auth local composable where the signup function is defined actually doesn't return the data from the signup request.

```
/node_modules/@sidebase/nuxt-auth/dist/runtime/composable/local/useAuth.mjs
const signUp = async (credentials, signInOptions, signUpOptions) => {
  const nuxt = useNuxtApp();
  const { path, method } = useTypedBackendConfig(useRuntimeConfig(), "local").endpoints.signUp;
  await _fetch(nuxt, path, {
    method,
    body: credentials
  });
  if (signUpOptions?.preventLoginFlow) {
    return;  // no data is returned
  }
  return signIn(credentials, signInOptions);
};
```

you'll notice that when `signUpOptions?.preventLoginFlow` is `true` like i have in my code  `{ preventLoginFlow: true }` it returns _zelch_ nothing.

And i believe this is the culprit.

### The Solution

Well I don't know if I'm savvy enough to implement the solution but from what i can see it looks like adding a variable to  hold the response from the `signup()`  `_fetch` request and then returning the variable with the data is the way to go.

for example

```
/node_modules/@sidebase/nuxt-auth/dist/runtime/composable/local/useAuth.mjs
const signUp = async (credentials, signInOptions, signUpOptions) => {
  const nuxt = useNuxtApp();
  const { path, method } = useTypedBackendConfig(useRuntimeConfig(), "local").endpoints.signUp;
  const response = await _fetch(nuxt, path, {   // added const response to hold fetch data
    method,
    body: credentials
  });
const data = response.json()  // parse the response body
  if (signUpOptions?.preventLoginFlow) {
    return data;   //return the data
  }
  return signIn(credentials, signInOptions);
};
```

[...]

PR #903, opened 2024-09-08
Enh(#843): Allow signup flow return data when preventLoginFlow is true

iamKiNG-Fr:
### 🔗 Linked issue

#843

### ❓ Type of change

### 📚 Description

Added functionality to allow user receive response from server after signup, when preventLoginFlow is true.

image
 

### 📝 Checklist


src/runtime/types.ts
```diff
@@ -53,6 +53,15 @@ export interface SessionDataObject {
     | SessionDataObject
 }
 
+/**
+ * Type of data returned from sign up functiom if preventLoginFlow is true
+*/
+export type SignUpResponse = {
```
zoey-kaiser, line 59: I think we should allow the passing of a generic type to the `signUp` function and then use that to type the response! 

This way we avoid hard-coding a type that may not fit to others `signUp` function and would allow them to pass a custom type!

zoey-kaiser (changes requested):

src/runtime/composables/local/useAuth.ts
```diff
@@ -154,17 +153,20 @@ async function getSession(getSessionOptions?: GetSessionOptions): Promise<Sessio
   return data.value
 }
 
-async function signUp(credentials: Credentials, signInOptions?: SecondarySignInOptions, signUpOptions?: SignUpOptions) {
+async function signUp<T = any>(credentials: Credentials, signInOptions?: SecondarySignInOptions, signUpOptions?: SignUpOptions): Promise<T> {
```
zoey-kaiser, line 156:
```suggestion
async function signUp<T>(credentials: Credentials, signInOptions?: SecondarySignInOptions, signUpOptions?: SignUpOptions): Promise<T> {
```

Can we remove the any typing here? This should result in the response being `unknown` if no type was passed.

zoey-kaiser (dismissed):

src/runtime/composables/local/useAuth.ts
iamKiNG-Fr: Sure, we can.

playground-local/server/api/auth/signup.post.ts
phoenix-ru, line 1: Could you please update your handler to be closer to a demo implementation introduced in #901?

https://github.com/sidebase/nuxt-auth/blob/main/playground-local/server/api/auth/login.post.ts

I think you either can:
1. return a "user" object and no access token (we are doing `preventLoginFlow`, right?) in https://github.com/sidebase/nuxt-auth/blob/5276a97997e4d1d5d3f0b76c0b064c0ab86e92cd/playground-local/server/api/auth/login.post.ts#L61-L65;
2. or do both "user" object and access tokens as in `signIn` handler so that reference is usable by others as inspiration and also by tests.
https://github.com/sidebase/nuxt-auth/blob/5276a97997e4d1d5d3f0b76c0b064c0ab86e92cd/playground-local/server/api/auth/login.post.ts#L67-L90

It is fine for me if you don't want to implement option 2, I can take it over.

playground-local/tests/local.spec.ts
```diff
@@ -59,4 +59,30 @@ describe('local Provider', async () => {
...
+    const page = await createPage('/register') // Navigate to signup page
+
+    const [
+      usernameInput,
+      passwordInput,
+      submitButton,
+    ] = await Promise.all([
+      page.getByTestId('regUsername'),
+      page.getByTestId('regPassword'),
+      page.getByTestId('regSubmit')
```
phoenix-ru, line 73: Unification
```suggestion
      page.getByTestId('username'),
      page.getByTestId('password'),
      page.getByTestId('submit')
```

src/runtime/composables/local/useAuth.ts
```diff
@@ -160,17 +159,20 @@ async function getSession(getSessionOptions?: GetSessionOptions): Promise<Sessio
...
+
+  // Holds result from fetch to be returned if signUpOptions?.preventLoginFlow is true
+  const result = await _fetch<T>(nuxt, path, {
     method,
     body: credentials
   })
 
   if (signUpOptions?.preventLoginFlow) {
-    return
+    // Returns result
```
phoenix-ru, line 174:
```suggestion
```

phoenix-ru (approved):

phoenix-ru (approved):

Issue #895, opened 2024-09-02
Custom refresh response token pointer


Rizzato95:
### Describe the feature

Hi, 

I'm using the module in `local` mode pointing to an external api (to which I don't have access and cannot edit).
During sign in the token comes in `bearerToken` property:

```typescript
{
  ...
  "bearerToken": "my-token",
}
```

But during refresh the token comes in `accessToken` property:

```typescript
{
  ...
  "accessToken": "my-token",
}
```

In the configuration I set `signInResponseTokenPointer` to `/bearerToken` who works like a charm during SignIn but returns an error during refresh:

![image](https://github.com/user-attachments/assets/eaf40ae0-bda5-4562-ad2a-889aaee50aae)

This happens because the module uses `signInResponseTokenPointer` to retrieve the token also during the refresh call.

What I need is a configuration property like `refreshResponseTokenPointer` to allow the token to be handled correctly when  provided in different properties between calls. 

Thanks

### Provider

- [ ] AuthJS
- [X] Local
- [ ] Refresh
- [ ] New Provider

phoenix-ru:
Hi @Rizzato95 , thanks for your request. As we're a bit busy at the moment with higher priority issues, would you give it a try yourself? We would be able to quickly review your PR

N1K1TAS95:
Same problem here. 
I'm coming from version 0.7.2, where there was that option.
I've also opened an issue #635 and it was fixed #727.

Rizzato95:
> Same problem here. I'm coming from version 0.7.2, where there was that option. I've also opened an issue #635 and it was fixed #727.

I think we're not talking about the same problem.
You asked for an option in the refresh request (which is now present as you correctly say), I need it in the refresh response. 

I created a PR (#910) for this new option.

N1K1TAS95:
@Rizzato95 ops 😅. You’re right, my bad.

phoenix-ru:
Closing as done in #910

PR #910, opened 2024-09-14
enh(#895): Custom refresh response token pointer

Rizzato95:
### 🔗 Linked issue

#895

### ❓ Type of change

* [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
* [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
* [x] 👌 Enhancement (improving an existing functionality like performance)
* [ ] ✨ New feature (a non-breaking change that adds functionality)
* [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
* [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Added `refreshResponseTokenPointer` option to allow the token assignment (during the `refresh` call) to a different property respect to the `signIn` call.

This is useful in cases where the server returns the token in a property with different names between `login` and `refresh`.

### 📝 Checklist

* [x] I have linked an issue or discussion.
* [ ] I have added tests (if possible).
* [x] I have updated the documentation accordingly.

### DIFF
--- a/docs/guide/local/quick-start.md
+++ b/docs/guide/local/quick-start.md
@@ -295,11 +295,13 @@ This follows the JSON Pointer standard, see its RFC6901 here: https://www.rfc-ed
 #### `refreshResponseTokenPointer`
 
 - **Type:** `string`
-- **Default:** `'/token'`
+- **Default:** `''`
 
 How to extract the authentication-token from the refresh response.
 
 E.g., setting this to `/token/bearer` and returning an object like `{ token: { bearer: 'THE_AUTH_TOKEN' }, timestamp: '2023' }` from the `refresh` endpoint will result in `nuxt-auth` extracting and storing `THE_AUTH_TOKEN`.
+
+If not set, `token.signInResponseTokenPointer` will be used instead.
 
 This follows the JSON Pointer standard, see its RFC6901 here: https://www.rfc-editor.org/rfc/rfc6901
 
--- a/docs/guide/local/quick-start.md
+++ b/docs/guide/local/quick-start.md
@@ -223,8 +223,8 @@ export default defineNuxtConfig({
         endpoint: { path: '/refresh', method: 'POST' },
         refreshOnlyToken: true,
         token: {
-          refreshResponseTokenPointer: '/token',
           signInResponseRefreshTokenPointer: '/refresh-token',
+          refreshResponseTokenPointer: '',
           refreshRequestTokenPointer: '/refresh-token',
           cookieName: 'auth.token',
           maxAgeInSeconds: 1800,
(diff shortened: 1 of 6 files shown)


docs/guide/local/quick-start.md
```diff
@@ -223,6 +223,7 @@ export default defineNuxtConfig({
         endpoint: { path: '/refresh', method: 'POST' },
         refreshOnlyToken: true,
         token: {
+          refreshResponseTokenPointer: '/token',
```
zoey-kaiser, line 226: Could we move it below the `signInResponseRefreshTokenPointer` inside the docs please 😊

docs/guide/local/quick-start.md
```diff
@@ -280,6 +281,17 @@ When refreshOnlyToken is set, only the `token` will be refreshed and the `refres
 
 ### `token`
 
+#### `refreshResponseTokenPointer`
```
zoey-kaiser, line 284: Same here

src/module.ts
```diff
@@ -76,6 +76,7 @@ const defaultsByBackend: {
       endpoint: { path: '/refresh', method: 'post' },
       refreshOnlyToken: true,
       token: {
+        refreshResponseTokenPointer: '/token',
```
zoey-kaiser, line 79: So personally I think we should apply the following logic for setting this value to ensure it does not result in breaking changes:

- if `token.refreshResponseTokenPointer` is set, we use it to retrieve the token
- if `token.refreshResponseTokenPointer` is not set, we use `signInResponseRefreshTokenPointer` to retrieve the token instead of setting a default value

This is primarily to avoid introducing a breaking change, as if someone set `signInResponseRefreshTokenPointer` to e.g. `/accessToken` in their setup, and uses a refresh token, then their refresh response at the moment also returns the same schema as their sign in request.

Also cc @phoenix-ru on your opinion

src/runtime/types.ts
```diff
@@ -243,6 +243,18 @@ export interface ProviderLocal {
...
+       *
+       * E.g., setting this to `/token/bearer` and returning an object like `{ token: { bearer: 'THE_AUTH_TOKEN' }, timestamp: '2023' }` from the `refresh` endpoint will
+       * result in `nuxt-auth` extracting and storing `THE_AUTH_TOKEN`.
+       *
+       * This follows the JSON Pointer standard, see it's RFC6901 here: https://www.rfc-editor.org/rfc/rfc6901
+       *
+       * @default /token  Access the `token` property of the refresh response object
+       * @example /       Access the root of the refresh response object, useful when your endpoint returns a plain, non-object string as the token
+       */
+      refreshResponseTokenPointer?: string
```
zoey-kaiser, line 257: Let's also move this one below the sign in token pointer

zoey-kaiser (changes requested):

src/module.ts
Rizzato95: I totally agree with you. 

Only one thing: when the `token.refreshResponseTokenPointer` is not set, we have to use `signInResponseTokenPointer` and not `signInResponseRefreshTokenPointer`. 
(You probably meant the same thing but just wrote the wrong option 😊)

zoey-kaiser: Haha, yes! Slight slipup from my side. You are completely right!

Rizzato95: Obviously with this new implementation the default value for `refreshResponseTokenPointer` has to be empty, do you agree?

zoey-kaiser: > Obviously with this new implementation the default value for refreshResponseTokenPointer has to be empty, do you agree?

Yup! We have a default set for the `signInResponseTokenPointer` which would be used as a fallback!

phoenix-ru (approved):

phoenix-ru:
@zoey-kaiser Could you please re-review this?

Issue #878, opened 2024-08-23
Load `baseURL` from runtime config


cngJo:
### Describe the feature

I usually have my API base URL for external APIs in:

```ts
  runtimeConfig: {
    public: {
      apiBaseUrl: 'https://my-awesome-backend.local'
    }
  },
```

and replicate it in `nuxt-auth` config

```ts
baseURL: process.env.NUXT_PUBLIC_API_BASE_URL ?? "https://my-awesome-backend.local",
```

But then I need the `NUXT_PUBLIC_API_BASE_URL` variable on build and runtime, which is a bit annoying. 

Therefore, it would be awesome to specify a JSON pointer to runtime config like it's done for the access / refresh tokens.

### Additional information

- [X] Would you be willing to help implement this feature?

### Provider

- [X] AuthJS
- [X] Local
- [X] Refresh
- [ ] New Provider

cngJo:
Maybe related to #797

phoenix-ru:
You can now set `baseURL` using

```ts
runtimeConfig: {
  public: {
    auth: {
      baseURL: 'https://example.com/your/base/url'
    }
  }
}
```

phoenix-ru:
Done in #913

PR #913, opened 2024-09-18
feat(#797, #878): set `baseURL` via environment variables and improve internal url detection

### LINKED ISSUE #797: RFC: Improve the handing of `baseURL` to be more intuitive

zoey-kaiser:
### 🔗 Linked issue

closes #797, #878

### ❓ Type of change

- [X] 👌 Enhancement (improving an existing functionality like performance)
- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📝 Checklist


zoey-kaiser:
## Functional tests:

### Only setting `auth.baseURL`

- `auth.baseURL` not set: 
  - Requests made to `/api/auth`
- `auth.baseURL` set to `http://localhost:8080`
  - Requests made to `http://localhost:8080/api/auth`
- `auth.baseURL` set to `http://localhost:8080/auth`
  - Requests made to `http://localhost:8080/auth`

### Combination of `auth.baseURL` and `AUTH_ORIGIN`

- `auth.baseURL` not set, `AUTH_ORIGIN` not set:
  - Requests made to `/api/auth`
- `auth.baseURL` set to `http://localhost:8080`, `AUTH_ORIGIN` not set:
  - Requests made to `http://localhost:8080/api/auth`
- `auth.baseURL` set to `http://localhost:8080`, `AUTH_ORIGIN` set to `http://localhost:8081`:
  - Requests made to `http://localhost:8081/api/auth`
- `auth.baseURL` set to `http://localhost:8080`, `AUTH_ORIGIN` set to `http://localhost:8081/auth`:
  - Requests made to `http://localhost:8081/auth`

playground-authjs/nuxt.config.ts
```diff
@@ -8,7 +8,7 @@ export default defineNuxtConfig({
     globalAppMiddleware: {
       isEnabled: true
     },
-    baseURL: `http://localhost:${process.env.PORT || 3000}`
+    baseURL: `http://localhost:${process.env.PORT || 3000}/api/auth`
```
phoenix-ru, line 11: This looks like a breaking change

phoenix-ru (commented):
In general I understand the intent and very like the testing comment you left. However, there are several points:

1. What happens to users who already use `baseURL` as e.g. `http://localhost:3000`? In my understanding, requests are now made to this URL without adding `/api/auth`? This is a breaking change and needs to have its own minor release + docs if we settle on it.
2. What happens to internal `$fetch` calls? Remember that they have a limitation - they only work when path starts with a `/`. And here are two facts that break it:
  1. When using `authjs` provider in which these calls are the most relevant, you need to set `AUTH_ORIGIN` or `baseURL` to a fully-specified URL (i.e. protocol, hostname, etc.);
  2. Calling a fully-specified URL using `$fetch` invokes _external_ fetch, meaning that a real HTTP call is made from the server to itself - and it has a relatively high cost, bottlenecking performance.

Any implementation that doesn't provide backwards compatibility with at least these two points is therefore highly discouraged :/
I'd like for us to get involved in an RFC discussion and come up with a spec of how calls are made depending on what variable is set. For example, we know for sure:
- `authjs` provider is the same Nuxt server and therefore should always prefer internal calls, regardless of the `AUTH_ORIGIN` (only taking into account `pathname`);
- `local` provider is isomorphic (with lean towards external backends, it seems), and origin needs to be taken literally - if it is set, we assume all calls are external.
With these in mind, it might even make sense to disconnect the implementations of two providers into separate URL utils.

phoenix-ru (approved):

shayr1:
I am using http://localhost:8000/api and then 
```
endpoints: {
        signIn: { path: '/auth/login', method: 'post' },
        signOut: { path: '/auth/logout', method: 'post' },
        signUp: { path: '/auth/register', method: 'post' },
        getSession: { path: '/auth/user', method: 'get' },
      }
```
I am not using `AUTH_ORIGIN`, just according to instructions above, however it is always using `http://localhost:3000`

I am using v0.10.0

PR #924, opened 2024-10-08
feat(#673): migrate to AuthJS

phoenix-ru:
### 🔗 Linked issue
#673

### ❓ Type of change

- [ ] ⚠️  Breaking change (fix or feature that would cause existing functionality to change)

### 📝 Checklist

### DIFF
--- a/src/runtime/composables/authjs/useAuth.ts
+++ b/src/runtime/composables/authjs/useAuth.ts
@@ -1,4 +1,4 @@
-import type { AppProvider, BuiltInProviderType } from 'next-auth/providers/index'
+import type { AppProvider, BuiltInProviderType } from '@auth/core/providers/index'
 import { defu } from 'defu'
 import { type Ref, readonly } from 'vue'
 import { appendHeader } from 'h3'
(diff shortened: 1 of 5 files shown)


src/runtime/composables/authjs/useAuth.ts
```diff
@@ -112,8 +112,12 @@ const signIn: SignInFunc<SupportedProviders, SignInResult> = async (provider, op
 
   const csrfToken = await callWithNuxt(nuxt, getCsrfToken)
 
-  const headers: { 'Content-Type': string, 'cookie'?: string | undefined } = {
+  const headers: { 'Content-Type': string, 'X-Auth-Return-Redirect': string, 'cookie'?: string | undefined } = {
     'Content-Type': 'application/x-www-form-urlencoded',
+    // This header is set to force AuthJS to return JSON
+    // https://github.com/nextauthjs/next-auth/blob/b86f7bb721e2fa4e9c3251b029095b0f50d46198/packages/next-auth/src/react.tsx#L270-L285
+    // https://github.com/nextauthjs/next-auth/blob/b86f7bb721e2fa4e9c3251b029095b0f50d46198/packages/core/src/index.ts#L160
+    'X-Auth-Return-Redirect': '1',
```
phoenix-ru, line 120: This is a new header which needs to be documented (or linked to `authjs` docs). We need to mention for users to add it to their header whitelists

phoenix-ru:
While testing the code in a private project, we unfortunately discovered that AuthJS itself errors on an existing database. Setting up a new project with the official Next starter results in the same error.

After an in-depth investigation, the issue lies with their new dependency `oauth4webapi` and the way it is used in new versions of AuthJS.

kogratte:
Hello there!

Thanks you very much for letting us aware of the ongoing work. 

From what I read, should I understand that the migration is now working for those not using a database? IE, we're using third party provider, aka keycloak. Should I expect the latest version to work as expected? If so, which version can I give a try?

phoenix-ru:
> now working for those not using a database?

Sorry for a misleading comment, I meant **code**base :fearful: 

The migration is simply not working - as stated, the issue is not on our side, but on the `authjs` side, namely the `oauth4webapi` they use. We tried their official Next.js starter (i.e. without this module and without Nuxt) and couldn't get it to work.

kogratte:
Hello @phoenix-ru! Thanks for working on this! Would it be possible to add a reference to the next issue?

phoenix-ru:
> Hello @phoenix-ru! Thanks for working on this! Would it be possible to add a reference to the next issue?

Hi @kogratte I may get back to it today if I have some time. We deprioritized the migration because we lost trust in `authjs` being stable (due to multiple times in the past as well) as we don't want to ship with a broken dependency

phoenix-ru:
Closing due to inactivity. This will probably be tried again this year with the newer `@auth/core` version

Issue #967, opened 2024-12-23
TypeScript complains if `signUp` endpoint gets disabled


root5427:
### Environment

------------------------------
- Operating System: Windows_NT
- Node Version:     v22.11.0
- Nuxt Version:     3.14.1592
- CLI Version:      3.17.1
- Nitro Version:    2.10.4
- Package Manager:  yarn@4.0.1
- Builder:          -
- User Config:      default
- Runtime Modules:  @primevue/nuxt-module@4.2.5, @nuxt/eslint@0.7.4, @nuxt/fonts@0.10.3, @sidebase/nuxt-auth@0.10.0
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/github-ti8md3c4?file=nuxt.config.ts%3AL11

### Describe the bug

In the documentation it says that `signUp` endpoint can be disabled by setting it to `false`, as shown in the example:

https://github.com/sidebase/nuxt-auth/blob/8d31537f51b14d90a4a7d703e83cb7e45f9b66c2/docs/guide/local/quick-start.md?plain=1#L62

![image](https://github.com/user-attachments/assets/e9d2d3a8-606c-4bf3-ad8a-74d5b8bb6cc2)

I get type error, when `signUp` endpoint is set to `false` in `local` provider options.

Error text:
```
Type 'false' has no properties in common with type '{ path?: string | undefined; method?: "head" | "get" | "patch" | "post" | "put" | "delete" | "connect" | "options" | "trace" | undefined; }'.ts(2559)
```

I assume that this happens because of this line of code:
https://github.com/sidebase/nuxt-auth/blob/8d31537f51b14d90a4a7d703e83cb7e45f9b66c2/src/runtime/types.ts#L93

PR #968, opened 2024-12-23
fix(#967): allow `signUp` endpoint to be disabled

root5427:
### 🔗 Linked issue
#967 

### ❓ Type of change

### 📚 Description
Documentation states that `signUp` endpoint may be disabled by setting it to `false`:
https://github.com/sidebase/nuxt-auth/blob/8d31537f51b14d90a4a7d703e83cb7e45f9b66c2/docs/guide/local/quick-start.md?plain=1#L54-L67

But if you try to do this, you will get the following typescript error in your IDE:
```
Type 'false' has no properties in common with type '{ path?: string | undefined; method?: "head" | "get" | "patch" | "post" | "put" | "delete" | "connect" | "options" | "trace" | undefined; }'. ts(2559)
```

And the case where `signUp` endpoint is disabled is not handled by the `signUp` function of the `useAuth` composable:
https://github.com/sidebase/nuxt-auth/blob/8d31537f51b14d90a4a7d703e83cb7e45f9b66c2/src/runtime/composables/local/useAuth.ts#L164-L178

This PR should fix these issues.

### 📝 Checklist

### DIFF
--- a/src/runtime/composables/local/useAuth.ts
+++ b/src/runtime/composables/local/useAuth.ts
@@ -5,8 +5,8 @@ import { jsonPointerGet, objectFromJsonPointer, useTypedBackendConfig } from '..
 import { _fetch } from '../../utils/fetch'
 import { determineCallbackUrl } from '../../utils/url'
 import { getRequestURLWN } from '../common/getRequestURL'
-import { formatToken } from './utils/token'
 import { ERROR_PREFIX } from '../../utils/logger'
+import { formatToken } from './utils/token'
 import { type UseAuthStateReturn, useAuthState } from './useAuthState'
 import { callWithNuxt } from '#app/nuxt'
 // @ts-expect-error - #auth not defined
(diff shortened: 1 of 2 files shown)


src/runtime/composables/local/useAuth.ts
```diff
@@ -163,8 +163,16 @@ async function getSession(getSessionOptions?: GetSessionOptions): Promise<Sessio
 
 async function signUp(credentials: Credentials, signInOptions?: SecondarySignInOptions, signUpOptions?: SignUpOptions) {
   const nuxt = useNuxtApp()
+  const runtimeConfig = useRuntimeConfig()
+  const config = useTypedBackendConfig(runtimeConfig, 'local')
+
+  const signUpEndpoint = config.endpoints.signUp
+
+  if (!signUpEndpoint) {
```
zoey-kaiser, line 171: Hey @root5427 👋 

Thanks for the contribution. I was wondering if we may want to add a `console.warn` message here that the endpoint is disabled? I feel like if someone disabled the endpoint by accident and then attempts to call the function regardless it would be nice to give a short warning to make it eaiser to find out why the function is not working 😄

zoey-kaiser (changes requested):

src/runtime/composables/local/useAuth.ts
root5427: Hi, Zoey.
Great suggestion, thank you very much. I will commit corresponding changes 👍

zoey-kaiser (approved):

zoey-kaiser (approved):
Approved! Thanks for the PR ❤️

I pushed a small lint fix, I hope that was all right. Happy holidays and have a good start into the new year! 🎉

PR #971, opened 2024-12-23
fix: Include cookie headers on session fetch

silverbackdan:
### 🔗 Linked issue

N/a

I can complete more if required, but I believe this to be a fix. When refreshing a local session from the client-side, the cookie headers are passed to the `getSession` endpoint. However on server-side requests these are missed off. This change introduces continuity.

The issue I faced was that I also have a session initialised with an identifier in a cookie. This identifier allows me to get the user session, and get a secure stored refresh token which is only available server-side to then refresh the user JWT token automatically. I can post this refresh token to my API to generate a new JWT token. But I need to be able to retrieve the session.

### ❓ Type of change

### 📝 Checklist


phoenix-ru:
I only have a question - are you using `getSession` on server-side? If yes, can you show how? [Docs](https://nuxt.com/docs/api/composables/use-request-headers) say that `useRequestHeaders` on browser returns an empty object.

phoenix-ru:
@silverbackdan Thank you for your PR! I recently started doing important patches for the `local` provider and now had significantly more context to understand the PR and be confident in its impact :slightly_smiling_face:

PR #1062, opened 2025-10-17
feat(#964): Add Hooks provider

### LINKED ISSUE #964: RFC: Hooks for the local provider
### Describe the feature

[Work in Progress]

It is an often requested feature to support hooking into the calls done by NuxtAuth. Common use-cases include:
- modifying the headers,
- modifying the request,
- reading data from the response

_No response_

### Additional information

- [X] Would you be willing to help implement this feature?

### Provider

- [ ] AuthJS
- [X] Local
- [X] Refresh
- [ ] New Provider

phoenix-ru:
### 🔗 Linked issue

Closes #964

### ❓ Type of change

### 📝 Checklist


docs/guide/hooks/adapter.md
```diff
@@ -0,0 +1,129 @@
...
+Must return either an object conforming to:
+
+```ts
+interface CreateRequestResult {
+  // Path to the endpoint
+  path: string
+  // Request: body, headers, etc.
+  request: NitroFetchOptions
+}
+```
```
zoey-kaiser: Why did you decide to abstract the returns of the hooks provider? While I think it is a good step forward compared to the current `local`-provider I think it could still limit you if you e.g. want to switch from `ofetch` to `axios` under the hood.

docs/guide/hooks/adapter.md
```diff
@@ -0,0 +1,129 @@
...
+
+or `false` to stop execution (no network call will be performed).
+
+### `authState` argument
+
+This argument gives you access to the state of the module, allowing to read or modify session data or tokens.
```
zoey-kaiser: Do we have a schema for this or this is e.g. based off the session object you can provide via the nuxt config?

docs/guide/hooks/adapter.md
```diff
@@ -0,0 +1,129 @@
...
+
+### `nuxt` argument
```
zoey-kaiser: Should we call this `nuxtApp` like Nuxt also does in the docs you linked below?

docs/guide/hooks/adapter.md
```diff
@@ -0,0 +1,129 @@
...
+
+This argument is provided for your convenience and to allow using Nuxt context for invoking other composables. See the [Nuxt documentation](https://nuxt.com/docs/4.x/api/composables/use-nuxt-app) for more information.
+
+## `onResponse(response, authState, nuxt)`
+
+Handle the response and optionally instruct the module how to update state.
+
+May return:
+* `false` — stop further processing (module will not update auth state).
```
zoey-kaiser: Does this `false` act the same way as the return `false` from the `createRequest` hook?

docs/guide/hooks/adapter.md
```diff
@@ -0,0 +1,129 @@
...
+* `undefined` — proceed with default behaviour (e.g., the `signIn` flow will call `getSession` unless `signIn()` options say otherwise).
+* `ResponseAccept` object — instruct the module what to set in `authState` (see below).
+* Throw an `Error` to propagate a failure.
+
+The `response` argument is the [`ofetch` raw response](https://github.com/unjs/ofetch?tab=readme-ov-file#-access-to-raw-response) that the module uses as well. `response._data` usually contains parsed body.
```
zoey-kaiser, line 110: What else can `response._data` contain if not the parsed body? E.g. do we do error handling here or just return whatever was in the body?

docs/guide/hooks/examples.md
```diff
@@ -0,0 +1,161 @@
...
+    onResponse(response) {
+      // Backend returns { access: 'xxx', refresh: 'yyy', user: {...} }
+      const body = response._data
+      // Default to `undefined` to not reset the tokens and session (but you may want to reset it)
+      return {
+        token: body?.access ?? undefined,
+        refreshToken: body?.refresh ?? undefined,
+        session: body?.user ?? undefined,
+      }
+    },
```
zoey-kaiser, line 33: Are these hooks promises? E.g. if I wanted to do another DB or API call here, would the hooks provider support this?

docs/guide/hooks/examples.md
```diff
@@ -0,0 +1,161 @@
...
+      return {
+        path: '/auth/profile',
+        request: {
+          method: 'post',
+          body: { token: authState.token.value },
+        },
+      }
+    },
+
+    onResponse(response) {
```
zoey-kaiser, line 53: If I return `false` from the `getSession.createRequest` hook is this `response`:

- `false`
- `null` / `undefined`
- `onResponse` is never called

docs/guide/hooks/adapter.md
zoey-kaiser: Do we want to add a guide on how you can use this to e.g. access `runtimeConfig` variables? I think this could be a pretty common workflow.

docs/guide/hooks/examples.md
phoenix-ru: Yes, the hooks are awaitable:
https://github.com/sidebase/nuxt-auth/blob/c416c1e972f2b1c8a62d7b7b7d3318bca1c69e30/src/runtime/composables/hooks/useAuth.ts#L94

docs/guide/hooks/examples.md
phoenix-ru: The `onResponse` is never called:
https://github.com/sidebase/nuxt-auth/blob/c416c1e972f2b1c8a62d7b7b7d3318bca1c69e30/src/runtime/composables/hooks/useAuth.ts#L73-L76

docs/guide/hooks/adapter.md
phoenix-ru: We do error handling for the request:
https://github.com/sidebase/nuxt-auth/blob/c416c1e972f2b1c8a62d7b7b7d3318bca1c69e30/src/runtime/composables/hooks/useAuth.ts#L78-L92

But thinking a bit more, we are silently consuming the error which is not very good for this usecase:
https://github.com/sidebase/nuxt-auth/blob/c416c1e972f2b1c8a62d7b7b7d3318bca1c69e30/src/runtime/utils/fetch.ts#L75-L88

It was done like so to avoid leaking exact error details (which might not be caught by consuming code and passed to the app side -> viewable by the frontend). I will consider if this could be improved, there's an external PR for better error handling:
https://github.com/sidebase/nuxt-auth/pull/1069

docs/guide/hooks/adapter.md
phoenix-ru: Yep, exactly the same - the meaning of `false` is to stop execution without an error

docs/guide/hooks/adapter.md
phoenix-ru: Yes, we do! I updated the documentation for it to point to [`useAuthState`](https://auth.sidebase.io/guide/application-side/session-access#useauthstate-composable):

image

docs/guide/hooks/adapter.md
phoenix-ru: Renamed

docs/guide/hooks/adapter.md
phoenix-ru: I don't think it would ever make sense for us to switch the underlying fetch library to something outside Nuxt ecosystem. At the moment `ofetch` is the default library of Nuxt, therefore it is being used

docs/guide/hooks/adapter.md
phoenix-ru: Will do, thanks

PR #1088, opened 2026-03-27
test: add authjs tests and bump dependencies

phoenix-ru:
### 🔗 Linked issue

N/A

### ❓ Type of change

### 📚 Description

This PR adds testing to `authjs` provider using Playwright, similar to what was done for `local` provider.

### 📝 Checklist


playground-local/package.json
```diff
@@ -12,17 +12,17 @@
...
+    "@nuxt/test-utils": "^4.0.0",
     "@playwright/test": "^1.58.2",
     "@types/node": "^24.10.11",
     "@vue/test-utils": "^2.4.6",
     "nuxt": "^3.20.2",
     "typescript": "^5.8.3",
-    "vitest": "^3.2.4",
-    "vue-tsc": "^2.2.12"
+    "vitest": "^4.1.2",
+    "vue-tsc": "^3.2.6"
```
zoey-kaiser: This is a pretty big bump to `vue-tsc`. Are there any other packages that make sense to update too (e.g. jump to `typescript` 6?)

src/runtime/composables/authjs/useAuth.ts
```diff
@@ -147,7 +147,7 @@ export function useAuth(): UseAuthReturn {
 
     const action: 'callback' | 'signin' = isCredentials ? 'callback' : 'signin'
 
-    const csrfToken = await getCsrfTokenWithNuxt(nuxt)
+    const csrfToken = await getCsrfToken()
```
zoey-kaiser, line 150: Why did you decide to remove the `WithNuxt` suffixes from the functions? From what I see we are still calling them inside the Nuxt context. 

Generally I would expect:
- Without the suffix: the regular function
- With the suffic: the function wrapped inside the `callWithNuxt` function

Please correct me if I am missing something here / overlooked some changes 😁

pnpm-lock.yaml
zoey-kaiser, line 1: Sanity check: did you run a dedupe on this?

phoenix-ru: Yes

src/runtime/composables/authjs/useAuth.ts
phoenix-ru: The idea is that functions were previously exported without `callWithNuxt`, so their users needed to manually wrap in `callWithNuxt` while it really was not necessary and we could export the context-bound functions immediately (since `useAuth` is a composable this makes more sense).
This might look like a change in the API behaviour, though

zoey-kaiser: Hmm okay. I would challenge if this change is needed inside a PR that aims to add more tests, as it would require at least a minor release then.

However, I am fine with proceeding like this 😄

playground-local/package.json
zoey-kaiser: Soooo? 👀

zoey-kaiser (dismissed):
Code approved! Can you please adjust the PR title to be more exact about the changes prior to merging 🙏

